PR #26, opened 2022-12-07
Unify docs content! No more Nitro submodule!

DZGoldman:
- content taken from https://github.com/OffchainLabs/nitro/commit/e38e7ef50cc86b482cd3bf677a37378ae0e8061f (except whitepaper, which will continue to live in `nitro`)
- note that edit url won't work in the preview, only when it's merged / live (since it takes you to master branch)
- upon merging (this PR), plz just merge, don't squash 🙏 

(see also https://github.com/OffchainLabs/nitro/pull/1372)


hkalodner (approved):
LGTM

PR #31, opened 2022-12-09
Update das docs with IPFS mirror instructions

Tristan-Wilson:
Also set default guidance for committee members to not use the local-disk-storage, and to enable discarding expired data on other storage types.

Update mirror examples to use
--data-availability.rest-aggregator.sync-to-storage.state-dir to avoid needing to resync from eager-lower-bound-block on restart.

### DIFF
--- a/arbitrum-docs/das/daserver-instructions.md
+++ b/arbitrum-docs/das/daserver-instructions.md
@@ -16,7 +16,7 @@ There are two main interfaces, a REST interface supporting only GET operations a
 
 Committee members listen on the REST interface and additionally listen on the RPC interface for `das_store` RPC messages from the sequencer. The sequencer signs its requests and the committee member checks the signature. The RPC interface also has a health check that checks the underlying storage that responds requests with RPC method `das_healthCheck`.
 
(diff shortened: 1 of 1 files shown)


arbitrum-docs/das/daserver-instructions.md
```diff
@@ -1,24 +1,38 @@
...
 
 Committee members listen on the REST interface and additionally listen on the RPC interface for `das_store` RPC messages from the sequencer. The sequencer signs its requests and the committee member checks the signature. The RPC interface also has a health check that checks the underlying storage that responds requests with RPC method `das_healthCheck`.
 
+IPFS is an alternative interface serving batch retrieval. A mirror can be configured to sync and pin batches to its local IPFS repository, then act as a node in the IPFS peer-to-peer network. A Nitro node that is configured to use IPFS that is syncing an AnyTrust chain that will use the batch hashes from L1 to find the batch data on the IPFS peer-to-peer network. Depending on network configuration that Nitro node may then also act as an IPFS node serving the batch data.
```
DZGoldman, line 19:
```suggestion
IPFS is an alternative interface serving batch retrieval. A mirror can be configured to sync and pin batches to its local IPFS repository, then act as a node in the IPFS peer-to-peer network. A Nitro node that is configured to use IPFS that is syncing an AnyTrust chain will use the batch hashes from L1 to find the batch data on the IPFS peer-to-peer network. Depending on network configuration, that Nitro node may then also act as an IPFS node serving the batch data.
```

DZGoldman (approved):
don't know if the config files and such need review (which I'm not the one for), but if not go ahead and merge @Tristan-Wilson

PR #53, opened 2022-12-20
...fix more broken links

DZGoldman:


fionnachan (approved):
That's unexpected from docusaurus 🥲

PR #56, opened 2022-12-24
Update why-nitro.mdx

clarus:


DZGoldman:
i'll let fine folks at cnbc [explain](https://www.cnbctv18.com/cryptocurrency/crypto-slang-wen-moon-rekt-rugged-flippening-nfa-14802321.htm) :)

clarus:
Ah OK interesting, thanks!

PR #58, opened 2022-12-27
Rename mainnet-beta.mdx to .arbitrum-mainnet-beta.mdx

Devjulius080:
The mainet of arbitrum under Ethereum Network


DZGoldman:
hi doesn't seem necessary (as its already under aribtrum-docs);(also all the link page-ids would need to be updated too)

PR #62, opened 2022-12-29
feat: add last updated time

fionnachan:
saw this config available on docusaurus
TBD: will it be useful?


DZGoldman (approved):
I like it; might be slightly misleading since any diff will trigger it, but I think it's a net good; lgtm

PR #75, opened 2023-01-15
Updating Local Dev Node information

saucepoint:
* Added some additional instructions for setting up this docs repo
* L2 nitro devnet runs on port 8547, not 7545
* Added information around the additional test wallet. The previous test wallet was only funded on L2 (and not L1)
    * sourced it from the [nitro script](https://github.com/OffchainLabs/nitro/blob/master/test-node.bash#L42)
* Provided some default addresses (sourced from [discord](https://discord.com/channels/585084330037084172/859511259183448084/1063583531089023116))

---

Not sure how often the test wallet private key changes and/or the default addresses, let me know if these shouldn't be hardcoded into the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,18 +4,24 @@ Arbitrum Docs, built with docusaurus; docs are live at https://developer.arbitru
 
 [Arbitrum SDK](https://github.com/OffchainLabs/arbitrum-sdk) auto-genned docs are included as submodule.
 
-## Local Build
-
-```shell
-cd ./website
-```
-
 ### Initial set up
 
+```shell
+git clone git@github.com:OffchainLabs/arbitrum-docs.git
+
+cd arbitrum-docs/
+
+git submodule update --init --recursive
+
+cd website/
+```
+
+Install node dependencies
 ```
 yarn
 ```
 
+You will only need to generate docs once:
 ```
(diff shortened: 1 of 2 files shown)


DZGoldman (changes requested):
Thank you @saucepoint! Small diff with ` yarn prettier:format`; run that and then lgtm!

saucepoint:
@DZGoldman should be good now!

Jason-Wanxt:
The docker container name of sequencer should be `nitro_sequencer_1` [here](https://github.com/saucepoint/arbitrum-docs/blob/local-dev-node-update/arbitrum-docs/node-running/local-dev-node.mdx#get-core-contracts) rather than `nitro-sequencer-1`, should we change it also?

saucepoint:
@Jason-Wanxt i think the name of the sequencer got changed at some point. getting the config via `nitro-sequencer-1` works for me (and the underscore version does not)

That reminds me, I remember seeing a log (from running `./test-node.bash`) that a script could not find the `nitro_sequencer_1` container

Jason-Wanxt (approved):
LGTM

mahsamoosavi (approved):
LGTM :)

DZGoldman (approved):

PR #76, opened 2023-01-16
Troubleshooting stubs

symbolpunk:
Stubs out the pages we need based on discussions on Slack [here](https://offchainlabs.slack.com/archives/C04ECJJ6VUH/p1673549984637669?thread_ts=1673546416.266979&cid=C04ECJJ6VUH). 

Notes:

 - I didn't use `*.mdx` extensions - we can actually use MDX components within vanilla markdown files; one benefit of this is that folks who have fresh installations of VS Code aren't required to configure anything in order to get markdown syntax highlighting. Only drawback I can think of is inconsistency between this and the existing norm of using `*.mdx`. Happy to pull up to norm ofc.
 
 - Tried my best to place these stubs in the file directories that match the sidebar structure. Possible benefit of this is that markdown files corresponding to docs are more easily findable.

 - We could introduce a `For Users` header link to bring us one step closer to identity-oriented header sections (eg `Users`, `Developers`, `Node runners`, `Researchers`, etc). Not yet aligned on this IA (alignment doc is [here](https://www.notion.so/arbitrum/FAQs-Glossaries-and-Info-Architecture-8ce21ee0e2a74212a6fe27903dd4bb88)). User-specific troubleshooting doc is stubbed here, not linked from a sidebar yet. Current idea is to stick user-specific troubleshooting into the root Troubleshooting doc until we have the `Users` header + docs.

 - Introduced YAML frontmatter as a possible standardization of the way that we align on metadata like content type, target audience, sme-owner, description (for SEO), title, etc. All of this could be pulled from Notion eventually
 
 - Next step is to fill it up with content pulled from Notion, coming soon 🎉


DZGoldman:
maybe for now (until we align on a broader restructuring) the "user troubleshooting" can be linked to under "intro" (since that's kind of the closing thing to be user-directed)?

DZGoldman (approved):

PR #97, opened 2023-02-06
adding lunr.js with contextual search

chrisco255:
Screen Shot 2023-02-05 at 8 07 43 PM

Adds contextual search with lunr.js as local search engine.

### DIFF
--- a/website/src/theme/SearchBar/index.js
+++ b/website/src/theme/SearchBar/index.js
@@ -1,0 +1,135 @@
+import React, { useRef, useCallback, useState, useEffect } from "react";
+import classnames from "classnames";
+import { useHistory } from "@docusaurus/router";
+import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
+import { usePluginData } from '@docusaurus/useGlobalData';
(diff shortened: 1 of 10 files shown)


website/src/theme/SearchBar/index.js
```diff
@@ -0,0 +1,135 @@
+import React, { useRef, useCallback, useState, useEffect } from "react";
```
DZGoldman, line 1: Maybe we include an inline comment with a link to the source of this code? (I believe [this](https://code.davidawindham.com/david/til/commit/6e35b1bbe2804bb17bf91b32cc77f77303fcf7a0))

DZGoldman (changes requested):

website/src/theme/SearchBar/index.js
DZGoldman: ...actually maybe I misunderstood - was this all just the content that was autogenned w/ swizzle? @chrisco255

chrisco255: yeah it was autogenerated from Swizzle but I did add a little but of custom CSS

DZGoldman (approved):
lgtm!

PR #101, opened 2023-02-07
Move link to latest nitro snapshot

ynvv:
Old snapshot link (https://snapshot.arbitrum.io/mainnet/nitro.tar) is very outdated. I think this snapshot is more recent.

### DIFF
--- a/arbitrum-docs/node-running/running-a-node.mdx
+++ b/arbitrum-docs/node-running/running-a-node.mdx
@@ -18,9 +18,9 @@ Note: If you’re interested in accessing an Arbitrum chain, but you don’t wan
 
 - Arbitrum One Nitro Genesis Database Snapshot
 
-  - Use the parameter `--init.url="https://snapshot.arbitrum.io/mainnet/nitro.tar"` on first startup to initialize Nitro database
-  - If running more than one node, easiest to manually download image from https://snapshot.arbitrum.io/mainnet/nitro.tar and host it locally for your nodes
-  - Or use `--init.url="file:///path/to/snapshot/in/container/nitro.tar"` to use a local snapshot archive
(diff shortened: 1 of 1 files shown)


PlasmaPower:
We should clarify that this database is a synced full database, so it's not usable for archive queries, which will require syncing from genesis (unless we host a synced archive database which I don't think we do?).

arbitrum-docs/node-running/running-a-node.mdx
```diff
@@ -18,9 +18,9 @@ Note: If you’re interested in accessing an Arbitrum chain, but you don’t wan
 
 - Arbitrum One Nitro Genesis Database Snapshot
 
-  - Use the parameter `--init.url="https://snapshot.arbitrum.io/mainnet/nitro.tar"` on first startup to initialize Nitro database
-  - If running more than one node, easiest to manually download image from https://snapshot.arbitrum.io/mainnet/nitro.tar and host it locally for your nodes
-  - Or use `--init.url="file:///path/to/snapshot/in/container/nitro.tar"` to use a local snapshot archive
+  - Use the parameter `--init.url="https://snapshot.arbitrum.io/mainnet/nitro-recent.tar"` on first startup to initialize Nitro database
+  - If running more than one node, easiest to manually download image from https://snapshot.arbitrum.io/mainnet/nitro-recent.tar and host it locally for your nodes
+  - Or use `--init.url="file:///path/to/snapshot/in/container/nitro-recent.tar"` to use a local snapshot archive
   - sha256 checksum: `a609773c6103435b8a04d32c63f42bb5fa0dc8fc38a2acee4d2ab2d05880205c`
```
chrisco255, line 24: If the file is different, the checksum should also be different.

DZGoldman:
ty @ynvv, but we're actually going to keep the old / smaller genesis snapshot as the recommend way to init sync

PR #102, opened 2023-02-09
Changes on Token Bridging page

TucksonDev:
- Added link to the StandardArbERC20 contract file in the Standard Bridging section, to follow the same style as the Generic-Custom gateway section.
- The two graphics of the Routers and the different types of Gateways are inside the Standard Bridging section and it would make more sense to have them in a more generic section, like the "Canonical Token Bridge Implementation" section, where we mention the different actors involved in the system.
- Added demo of the custom-token-bridging in the Demos section
- Changed title level of Other Flavors of Gateways as it should be at the same level as Standard and Generic Custom bridging sections
- Have "A word of Caution on Bridges" in its own section


mahsamoosavi (approved):
LGTM :)

PR #105, opened 2023-02-16
DAS RPC no longer supports getbyhash, update doc

Tristan-Wilson:
This method is still available on the REST endpoint.


PlasmaPower (approved):
LGTM

PR #116, opened 2023-02-23
Document diffs in JSON-RPC responses between mainnet and Arb

chrisco255:
## Overview

This PR attempts to improve [Arbitrum/Ethereum Differences](https://developer.arbitrum.io/arbitrum-ethereum-differences) doc by noting specific fields in JSON-RPC responses for transactions, blocks, and transaction receipts and how they sometimes differ from what is returned from a mainnet Ethereum RPC.

## Questions

- [ ] Are there any diffs between logs (as returned by `eth_getLogs` and other related methods)?

- [ ] Is there anything special to note regarding Arb Classic transactions, blocks, etc that should be noted here?


website/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
DZGoldman, line 1: vat is this

DZGoldman: ...more actionably, can we move this (and the yarn.lock diff) to its own PR?

DZGoldman (changes requested):
will review content, one orthogonal request

arbitrum-docs/arbitrum-ethereum-differences.mdx
```diff
@@ -44,3 +37,66 @@ Of particular note is the [ArbAddressTable](./arbos/precompiles#ArbAddressTable)
...
+#### Transaction Types
+
+In addition to the [three transaction types](https://ethereum.org/en/developers/docs/transactions/#types-of-transactions) currently supported on mainnet, Arbitrum adds additional types listed below and [documented in full detail here](arbos/geth#transaction-types).
+
+On calls that return custom Arbitrum type transactions, such as [`eth_getTransactionByHash`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionbyhash), the `type` field will reflect the custom codes where applicable. 
+
+- `100` - `ArbitrumDepositTxType`: A deposit of ETH from L1 to L2 via the Arbitrum bridge
+- `101` - `ArbitrumUnsignedTxType`: An L1 user can use to call an L2 contract via the bridge
+- `102` - `ArbitrumContractTxType`: Allows an L1 contract to call an L2 contract method via the bridge
+- `104` - `ArbitrumRetryTxType`: Used to [redeem a retryable ticket](arbos/l1-to-l2-messaging#redeeming-a-retryable) on L2, which finalizes a L2 to L1 message that failed to redeem automatically (usually due to low gas)
```
DZGoldman, line 81:
```suggestion
- `104` - `ArbitrumRetryTxType`: Used to [redeem a retryable ticket](arbos/l1-to-l2-messaging#redeeming-a-retryable) on L2, which finalizes a L1 to L2 message that failed to redeem automatically (usually due to low gas)
```

arbitrum-docs/arbitrum-ethereum-differences.mdx
```diff
@@ -44,3 +37,66 @@ Of particular note is the [ArbAddressTable](./arbos/precompiles#ArbAddressTable)
...
+- `105` - `ArbitrumSubmitRetryableTxType`: Retryable tickets are [submitted via the L1 bridge](/arbos/l1-to-l2-messaging#submitting-a-retryable) and allow arbitrary L1 to L2 messages to be created and executed on L2
```
DZGoldman, line 82: Something to think about (prob more generally than in this PR, but maybe nothing actionable here) is to distinguish between L1 to L2 messages generally and retryables in particular. I.e., "L1 to L2 message" could be an L2 message created b an L1 contract and sent directly to the slow inbox; what distinguishes retryables is submision and execution happen separately

DZGoldman (changes requested):
tysm, looks great, just little things

website/.yarnrc.yml
chrisco255: This is something Yarn v3 uses for its local node_modules cache, which is a folder of zip files of the project's dependencies that can be optionally checked into source control. I imagine some people use this feature for dependable static builds without needing to fetch dependencies from npm, but I added the `.yarn` cache to the `.gitignore` in a previous commit, so that only shows up locally.

Apparently I'm the only one that uses Yarn v3, so this hasn't come up before. I'll pull these changes out into another PR.

arbitrum-docs/arbitrum-ethereum-differences.mdx
chrisco255: Good call out, yeah, definitely something that would be better to clearly distinguish in other places in the docs

arbitrum-docs/arbitrum-ethereum-differences.mdx
```diff
@@ -44,3 +37,66 @@ Of particular note is the [ArbAddressTable](./arbos/precompiles#ArbAddressTable)
...
+- `from`: For L1-to-L2 related transactions, will be the _aliased_ version of the L1's `msg.sender`. See [L1 to L2 Messaging](./arbos/l1-to-l2-messaging.mdx) for more info. 
+
+See full list of [mainnet transaction fields](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionbyhash) for additional reference.
+
+### Tx Receipts
+
+#### Additional Fields
+
+- `l1BlockNumber`: The L1 block number that would be used [for block.number calls](time).
+- `gasUsedForL1`: Amount of gas spent on L1 computation in units of L2 gas.
```
DZGoldman, line 100: sry one more thing - i think this is a bit more accurate / clear:
```suggestion
- `gasUsedForL1`: Amount of gas spent on L1 calldata in units of L2 gas.
```

DZGoldman (changes requested):

DZGoldman:
@cla-bot check

DZGoldman (approved):
lgtm!

PR #118, opened 2023-02-26
Added Alchemy

vardhanshorewala:
Hey there - Added Alchemy's Ethereum development framework as an additional option for developers. Have used it in the past and is extremely lightweight and easily deployable.


DZGoldman (changes requested):
hi @vardhanshorewala — I think it would make more sense to link directly to the sdk, maybe https://docs.alchemy.com/reference/alchemy-sdk-quickstart is the best link? what do you think?

vardhanshorewala:
hey @DZGoldman — i agree with your suggestion as it makes it more specific. thanks!

DZGoldman:
@vardhanshorewala great — would you mind pushing the update + signing our CLA? then LGTM!

DZGoldman:
@cla-bot check

DZGoldman (approved):

PR #119, opened 2023-02-26
Added Link to Additional Dev Tools and dApps

vardhanshorewala:
Hey there - added Alchemy’s dApp store as I’ve found it to be a useful web directory. It has information about popular dApps and their respective chains. Thought this would help other devs directly view and play around with productionized dApps, enhancing their understanding from the rest of the article.


DZGoldman:
@cla-bot check

symbolpunk:
Thanks for the PR. This link is nice to have, but the purpose of this particular document is to help readers bridge tokens, and the addition of this link doesn't directly support that objective.

We'll soon have pages dedicated to collating third-party resources like this, so this is useful for future reference. Closing.

PR #126, opened 2023-03-01
Fix broken URL in optional parameter

0xpanoramix:
## Description

While I was trying to run an archive node, I had to use the `init.url` parameter using the given value on the documentation.
But the presence of the double-quotes prevent the node to start because the URL could not be parsed.
This PR removes those double-quotes.


DZGoldman (approved):
ty @0xpanoramix! if you could sign the CLA, then lgtm

DZGoldman:
@cla-bot check

DZGoldman (approved):
lgtm

PR #127, opened 2023-03-06
fix: sequencer should submit tx onto L1

Jason-Wanxt:


DZGoldman:
both are kinda true 🤷

Jason-Wanxt:
Thought submit here should mean send tx to contracs rather than feed tx.

DZGoldman (approved):

PR #136, opened 2023-03-10
Update the path of Nitro database for faster sync

BohemiaLai:


TucksonDev:
Thank you @BohemiaLai, but we prefer to keep the old / smaller genesis snapshot as the recommended way to init sync.

PR #171, opened 2023-03-23
feat: updated announcement banner post token launch

dewanshparashar:


bartlomiej94 (approved):

DZGoldman (approved):

DZGoldman (changes requested):
i think we should still link to the docs actually, wa u think @dewanshparashar

PR #229, opened 2023-04-21
fix: rename to staker in validator config

gzeoneth:
https://github.com/OffchainLabs/nitro/releases/tag/v2.0.12
> All --node.validator.* options were renamed to --node.staker.*

### DIFF
--- a/arbitrum-docs/node-running/running-a-validator.mdx
+++ b/arbitrum-docs/node-running/running-a-validator.mdx
@@ -22,10 +22,10 @@ Here we describe different types of validators and provide instructions on how t
 
 ### Running Watchtower Validator
 
-- Running a validator in `Watchtower` mode is the same as running a Nitro node with `--node.validator.enable --node.validator.strategy=Watchtower`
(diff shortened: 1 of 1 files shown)


fionnachan (approved):

arbitrum-docs/node-running/running-a-validator.mdx
```diff
@@ -22,10 +22,10 @@ Here we describe different types of validators and provide instructions on how t
 
 ### Running Watchtower Validator
 
-- Running a validator in `Watchtower` mode is the same as running a Nitro node with `--node.validator.enable --node.validator.strategy=Watchtower`
+- Running a validator in `Watchtower` mode is the same as running a Nitro node with `--node.staker.enable --node.staker.strategy=Watchtower`
```
fionnachan, line 25: what happens if they try to use the old param? do they get a prompt saying it's deprecated but it would still work?

it'd be better to add a short message saying from which version has it been replaced

gzeoneth: no it just say invalid flag

PR #260, opened 2023-05-04
Add support for variables

TucksonDev:
Attempt to add support for variables within the docs. This PR presents the basic structure and a use case: node images

Current behavior is:
- Any text, code or URL (in link tags) enclosed in @ characters can be replaced by the correspondent variable (e.g. @latestNitroVersion@)
- Variables are kept in `website/static/globalVars.js`

There are some limitations for this support:
- We can use other pattern, but if we want to support injecting variables in backticks enclosed texts, we can't use the following characters: [, {, (, |, &, = (probably some more). These characters are handled differently by docusaurus when converting those texts to "code" tags, to support syntax highlighting.
- Variables inside single backticks are not detected, but those can be replaced by < code> HTML tags (the behavior is the same)

### DIFF
--- a/website/src/remark/variable-injector.js
+++ b/website/src/remark/variable-injector.js
@@ -1,0 +1,57 @@
+// Based on https://github.com/facebook/docusaurus/issues/395
+// For future reference, we can virtually use any pattern to match agains variables,
+// but there's a limitation for handling text enclosed between single backticks (`) or triple backticks (```)
+// Docusaurus formats the code enclosed in backticks to apply certain styles to some characters
+// (to highlight instructions, operators, ...)
+// When using those characters as part of the pattern, they will not match and the variable will not be
+// injected.
+// These characters are: [, {, (, |
+// Here are some valid patterns
+//    ^FOO => /\^([a-zA-Z0-9_-]+)/g
+//    ::FOO => /\:\:([a-zA-Z0-9_-]+)/g
+//    @ FOO @ => /\@\s*([a-zA-Z0-9_-]+)\s*\@/g
+//    :: FOO :: => /\:\:\s*([a-zA-Z0-9_-]+)\s*\:\:/g
+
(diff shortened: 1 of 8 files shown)


mahsamoosavi (approved):
This will make our lives much easier! 🧚🏻‍♀️

website/src/remark/variable-injector.js
```diff
@@ -0,0 +1,57 @@
...
+//    @ FOO @ => /\@\s*([a-zA-Z0-9_-]+)\s*\@/g
+//    :: FOO :: => /\:\:\s*([a-zA-Z0-9_-]+)\s*\:\:/g
+
+const visit = require('unist-util-visit');
+
+const plugin = (options) => {
+  const transformer = async (ast) => {
+    visit(ast, ['text', 'code', 'link'], (node) => {
+      // Replace all occurrences of [[ varName ]] with the value of varName
+      let value;
```
DZGoldman, line 21: could use iffy syntax here (no necessary tho)

website/src/remark/variable-injector.js
```diff
@@ -0,0 +1,57 @@
...
+          value = node.url;
+          break;
+        
+        case "text":
+        case "code":
+          value = node.value;
+          break;
+      }
+
+      value = value.replace(/\@\s*([a-zA-Z0-9_-]+)\s*\@/g, (match, varName) => {
```
DZGoldman, line 33: comment explaining in words? (I think just "stuff between two "@"s)?

DZGoldman (changes requested):
pumped

TucksonDev:
@DZGoldman , I forgot to revisit that file after finishing with the rest of changes. I've added more clarification and removed outdated comments :)

symbolpunk (approved):

DZGoldman (approved):
lgtm

PR #292, opened 2023-05-24
Local build fix

symbolpunk:
- When building docs locally on Win10/Win11, builds fail
 - Builds don't fail when building locally on MacOS
 - I was able to isolate the issue to the `arbitrum-sdk` plugin configuration in `docusaurus.config.js` - commenting this out gets builds working on Win-local
 - commenting out code / adding `docusaurus.config.js` to `.gitignore` isn't clean and requires constant "remembering"; this hack solves the problem in a way that keeps Windows contributors unblocked without having to think about this (sdk gen still works locally & remotely; ci/cd isn't impacted)


TucksonDev (approved):
LGTM

PR #297, opened 2023-05-26
Update debian-11.6.0-amd64-netinst.iso download link

seonggwonyoon:
Since the version (11.6.0) has been remove from the official website, replace it with the iso from archive.org


TucksonDev:
Thanks for submitting this PR!
This change has been handled here: #355

PR #325, opened 2023-06-05
Fix typo

TomTaehoonKim:
I fixed a typo where the numbers don't match with the figure above.


TomTaehoonKim:
I've signed the document can you rerun it please?

TucksonDev (approved):
Thanks for fixing the typo!

PR #335, opened 2023-06-08
Improve text within the node running quickstart page

mahsamoosavi:
- Adds consideration for when one may want to run a classic node
- Changes some texts across the run-a-node subpages


TucksonDev (approved):
LGTM

PR #378, opened 2023-06-28
Improve local dev node how-to page

TucksonDev:
Added more useful information that is now also available in the [repo](https://github.com/OffchainLabs/nitro-testnode/).

[Preview](https://nitro-docs-git-improve-local-dev-node-page-offchain-labs.vercel.app/node-running/how-tos/local-dev-node)


arbitrum-docs/node-running/how-tos/local-dev-node.mdx
```diff
@@ -37,18 +33,51 @@ To relaunch the node after the first installation, run the following command.
...
+
+You can find a list of additional arguments to use with `test-node.bash` by using `--help`.
+
+```bash
+./test-node.bash --help
+```
+
+### Helper scripts
+
+The repository includes a set of helper scripts for basic actions like funding accounts or bridging funds. You can see a list of the available scripts by running:
```
tsahee, line 46: I think the one script most users would use is funding an address. I suggest adding a direct example for that here that will require minimal reading/expirementation to fit to their needs.

should be something like

```
./test-node.bash script send-l2 --to address_0x11223344556677889900 --ethamount 5
```

tsahee (approved):
Seems o.k. I suggested a small addition in comments.

arbitrum-docs/node-running/how-tos/local-dev-node.mdx
TucksonDev: Great idea. Added.
Thanks!

Issue #123, opened 2023-02-27
use prettier config in monorepo


DZGoldman:
https://github.com/OffchainLabs/config-monorepo/tree/main/packages/prettier-config

PR #394, opened 2023-07-10
Config prettier like monorepo and format almost-trivial files

TucksonDev:
This PR updates the configuration of prettier to make use of the [config-monorepo](https://github.com/OffchainLabs/config-monorepo/blob/main/packages/prettier-config/README.md). It also includes most of the files that need formatting. There are still some files that need reformatting, but those require a deeper analysis, so they'll be included in a future PR.

Prettier formats almost everything as expected. The only thing that doesn't allow to configure is the use of asterisks instead of underscores for italics and bolds. It feels more natural to me to use asterisks, but afaik there's no option on prettier that allows us to use that, or to even ignore those.


TucksonDev:
Closes #123

mahsamoosavi (approved):
So much prettier!💄

PR #396, opened 2023-07-12
Format files with more complex changes

TucksonDev:
This PR formats the rest of files in the repo. These are files that suffered a fair amount of formatting changes. They also had a few markdown inconsistencies that made prettier format them in an unexpected way, so I fixed the inconsistencies before running the formatter.

A few notes about the changes:
- Some of the files that were formatted before have now changes because I adjusted some of the prettier options related to JSX
- There are `{' '}` characters in some of the files that prettier uses to add spaces before or after "code" tags when they are in different lines. I haven't found a way to prevent this behaviour and rearranging the lines doesn't work either. This happens in l1-to-l2-messaging and orbit-gentle-introduction files.
- I also changed the scripts to run with yarn to remove the "prettier" word from them


symbolpunk (approved):
LGTM (aside from outstanding conflict res) 👍

PR #402, opened 2023-07-14
Improve SDK documentation

TucksonDev:
This PR changes the process by which the SDK documentation is generated. It transfers the process to this repo (instead of running the command in the sdk repo), uses and updated version of the markdown plugin for typedoc, and improves the general output by reestructuring the information displayed and generating a custom sidebar.

There are several important changes that should be reviewed:
- Upgraded typedoc-plugin-markdown version to the latest "next" version because it contains many improvements over the latest released version. To avoid breaking future deployments if a new "next" version is released, I've hardcoded the one being used (4.0.0-next.17). You can see the difference between the [current stable version](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/typedoc-plugin-markdown%403.15.0/packages/typedoc-plugin-markdown) and the ["next" version](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/next/packages/typedoc-plugin-markdown) in those links.
- typedoc generated documentation and docusaurus are highly dependent on each other, so I've installed typedoc and its dependencies in this repo. In any case, we can revert back to have those in the SDK repo if needed (we would just need to add the new config there)
- Modified some of the package.json scripts to handle typedoc in this repo instead of the SDK
- Documentation is now generated in `website/sdk-docs` instead of arbitrum-sdk/docs. It makes sense tonot place documentation files there. I'm not sure if "website" is the best place to add those documents to, but looks like a good option. I'm open for alternative suggestions.

[Preview](https://nitro-docs-git-improve-sdk-output-offchain-labs.vercel.app/sdk)


DZGoldman (approved):
lgtm!

PR #434, opened 2023-08-02
Create github action workflow with prettier check and reactivate broken links check on build

TucksonDev:
This PR does the following:
- Reformat files that were not currently formatted
- Add a github action for checking that files are formatted
- Reactivate broken links detection on both dev and prod builds: up to now, docusaurus would `warn` on dev and would `ignore` on prod. It now `throws` on both

By merging this PR, two changes will apply for next contributions:
1. We will have to run `yarn format` before updating the PR, otherwise the github check will fail
2. We will have to verify that there are no broken links by running yarn build or yarn start, otherwise the Vercel deployment will fail


symbolpunk (approved):
🔥

PR #505, opened 2023-08-31
Glossary updates: stylus things, parent/child chain nomenclature

DZGoldman:


TucksonDev (approved):
LGTM. Only flagging that "Parent chain" and "Child chain" are mostly written with the first letter in uppercase. Should we write them in lowercase?

DZGoldman:
@TucksonDev I don't see a way in notion to modify the display text in a glossary link ref (tho maybe I'm missing it?)

TucksonDev:
Oh, that's right 🤔 . Let's leave it like that then. If we find a good way to change the rendering in the notion lib, we can do that later (might be a bit tricky though).
Thanks!

PR #528, opened 2023-09-04
Add JSX nodes to variable injector

### LINKED ISSUE #515: global var replacement doesn't work for inline html content
Seems if the line  with html begins with text, the text node gets recognized, and it renders; otherwise it doesn't. i.e.,

```
this works @nitroRepositorySlug@
this also does <p> @nitroRepositorySlug@ </p>
<p>this doesn't @nitroRepositorySlug@ </p>
```

TucksonDev:
Fixes #515 

Adds JSX AST nodes to the variable injector plugin.


DZGoldman (approved):
ty, lgtm!

PR #532, opened 2023-09-06
Update testnet-information.md

amarrazza:


TucksonDev (approved):
🙏

hkalodner (commented):
For "there is no 7-day delay, withdrawals are processed in minutes", I think we should be more clear that the reason for not having a delay is because it's a testnet, not because there's anything magic with stylus

hkalodner (approved):
LGTM

PR #533, opened 2023-09-06
Update FAQs format, add frontmatter and add script to detect orphan pages

TucksonDev:
This PR cleans part of the code of the automatically generated FAQs, so they have a proper frontmatter (and generate all HTML meta tags), and they have consistent styles with the rest of pages.

It also includes a script to detect orphan pages. Orphan pages are still present in the structure, but we can detect them now (this should be in another PR, but somehow mixed branches and ended up here).


arbitrum-docs/partials/_troubleshooting-building-partial.md
```diff
@@ -5,7 +5,7 @@
 
 <p>The L1 component is meant to compensate the Sequencer for the cost of posting transactions on L1 (but no more). (See <a href="https://developer.arbitrum.io/arbos/l1-pricing">L1 Pricing</a>.)</p>
 
-<p>The L2 component covers the cost of operating the L2 chain; it uses Geth for gas calculation and thus behaves nearly identically to L1 Ethereum. One difference is that unlike on Ethereum, Arbitrum chains enforce a gas price floor; currently 0.1 gwei on Arbitrum One and 0.01 gwei on Nova (See [Gas](/arbos/gas.mdx)).</p>
```
symbolpunk, line 8: non-blocking nit - should we normalize relative links over absolute links wherever possible?

arbitrum-docs/partials/_troubleshooting-users-partial.md
```diff
@@ -85,13 +85,15 @@
 
 <p>The <em>Validators</em> are the ones responsible for the safety of the chain; i.e., making staked claims about the chain state, disputing each other, etc.</p>
 
-<p>Currently, on Arbitrum One, the Sequencer is a centralized entity maintained by Offchain Labs. Eventually, the single Sequencer will be replaced by a distributed committee of Sequencers who come to consensus on transaction ordering. This upgrade will be an improvement; we don't want you to have to trust us not to reorder your transactions. However, it also isn't <em>strictly</em> necessary for Arbitrum One to achieve its most fundamental properties.</p>
+<p><a href="https://docs.arbitrum.foundation/state-of-progressive-decentralization">Currently</a>, on Arbitrum One, the Sequencer is a centralized entity maintained by Offchain Labs. Eventually, we expect the single Sequencer to be replaced by a distributed committee of Sequencers who come to consensus on transaction ordering. This upgrade will be an improvement; we don't want you to have to trust us not to reorder your transactions. However, it also isn't <em>strictly</em> necessary for Arbitrum One to achieve its most fundamental properties.</p>
```
symbolpunk, line 88: Non-blocking pattern nit - readers might hover over this link to see where it points, but others might not click it because they're not sure where it leads. Definitely a balance to strike between being _concise_ vs _unambiguous and inviting_... 

Something like "visit x doc to learn more" can help (and _might_ also help a small bit with SEO, by adding the page title to the text - but I'm not confident about that)

symbolpunk:
(just noticed the `Draft` tag - sorry for jumping the gun!)

arbitrum-docs/partials/_troubleshooting-building-partial.md
TucksonDev: Agree. Links in Notion sourced content are more complicated to deal with, because they don't render if they are not actual links in Notion, so the relative links don't work. I'm already planning a revamp of the library to render links as relative (wherever possible), and render actual markdown instead of HTML here. I did a few tests and it shouldn't be too difficult to do.

arbitrum-docs/partials/_troubleshooting-users-partial.md
TucksonDev: 100% agree. This one I hadn't notices. I'll keep it in mind for the FAQs revamp.

arbitrum-docs/for-devs/troubleshooting-building.md
```diff
@@ -1,11 +1,11 @@
...
+sidebar_label: 'Troubleshooting: Building Arbitrum dApps'
+description: List of questions and answers frequently asked by developers
+reader_audience: developers who want to build on Ethereum/Arbitrum
+content-type: faq
 ---
 
-import FaqPartial, { toc as FAQTOC } from '../partials/_troubleshooting-building-partial.md';
+import FaqPartial from '../partials/_troubleshooting-building-partial.md';
 
-<div data-faq-origin-slug="build-faq">
```
TucksonDev, line 7: @symbolpunk , you might notice I removed this structure as I didn't see it being used anywhere. Can you confirm that we can simplify that structure as I did?
Thanks! 🙏

symbolpunk: LGTM!

FYI, the original design assumed that we'd

 1. Generate a mono-FAQ that contains all FAQs on Notion
 2. Embed the entire mono-FAQ on sub-pages
 3. Use `data-faq-origin-slug` to hide irrelevant FAQs via CSS
 
 ...the idea being that we can include/exclude FAQs from pages by adding/removing slug tags on Notion instead of dealing with partial markdown files. But I don't think we leaned into this approach, and I don't think we need to :).

arbitrum-docs/partials/_troubleshooting-users-partial.md
```diff
@@ -85,13 +85,15 @@
...
 
 <p>In other words: <em><strong>An Arbitrum Rollup chain with a centralized Sequencer could theoretically still be trustless!</strong></em></p>
 
 <p>Which is to say — the more important thing than decentralizing the Sequencer, i.e., the thing you ought to care more about — is decentralizing the <em>Validators</em>.</p>
 
-<p>Arbitrum One's Validator set is currently allowlisted; overtime, the allowlist will grow and then be removed entirely. For more info see <a href="https://developer.arbitrum.io/mainnet-beta">"Mainnet Beta"</a>.</p>
+<p>Arbitrum One's Validator set is currently whitelisted; overtime, we expect <a href="https://docs.arbitrum.foundation/">governance</a> to expand the whitelist and eventually be removed entirely. For more info see <a href="https://docs.arbitrum.foundation/state-of-progressive-decentralization">"State of Progressive Decentralization"</a>.</p>
```
symbolpunk, line 94: "allowlist" was selected for a [good reason](https://www.notion.so/Why-don-t-we-say-blacklist-or-whitelist-anymore-fbd38356791c41d2a9f1dbbd0c2f919e); IMO we should stick to it.

Source doc [here](https://www.notion.so/arbitrum/Are-Sequencers-the-same-entities-as-Validators-Can-a-centralized-Sequencer-act-maliciously-e-g-875d764e5f3546ce9bc11a6427159c2a)

symbolpunk (changes requested):

arbitrum-docs/partials/_troubleshooting-users-partial.md
TucksonDev: 100%. I just changed the FAQ and also some other mentions to "whitelist" in the docs.
Thanks for flagging! 🙏

arbitrum-docs/partials/_troubleshooting-users-partial.md
TucksonDev: I ended up changing the FAQ because we already had a link to the same page at the end of the question, so it makes sense to leave that word without the link. 
Thanks!

arbitrum-docs/for-devs/troubleshooting-building.md
TucksonDev: Understood. Yeah, we are already filtering out the FAQs in Notion, and it might make sense the have the filter there (in Notion).
If we ever change the system, we can add the wrappers back.
Thanks!

symbolpunk (approved):

PR #602, opened 2023-09-21
FAQ discoverability, sidebar title conventions

symbolpunk:


mahsamoosavi (approved):

TucksonDev (approved):
LGTM, thanks @symbolpunk ! 🚀 

Something to think about for later: the FAQs page in "Developers" and "Users" is called "Troubleshooting", while we use "Frequently Asked Questions" in "Node runners" and "How it works". Should we follow the same pattern in all sections?

symbolpunk:
@TucksonDev Thank you!

> Should we follow the same pattern in all sections?

Yes :)

Our [current pattern guidance](https://www.notion.so/arbitrum/Pattern-guide-Docs-dafaf8d1ce34433d88c1ef3b4b99c19c?pvs=4#2309f7f5db284126929065ac3c56fef8) defines _Troubleshooting_ and _FAQ_ pages as distinct content types, each with its own purpose.

The current Dev Troubleshooting page looks like a FAQ, but it's meant to contain questions specifically oriented around troubleshooting as a starting point.

As a possible next step, we could refactor it to align with the "problem-solution mapping" purpose, similar to the way [Scenarios and Solutions are presented in Prysm's docs](https://docs.prylabs.network/docs/troubleshooting/issues-errors#troubleshooting-scenarios-and-solutions).

Then if/as support request volume increases, we can introduce the troubleshooting UX we implemented in the [Node troubleshooting page](https://docs.arbitrum.io/node-running/troubleshooting-running-nodes).

PR #627, opened 2023-09-28
add arbitrum ecosystem graphic to nodes providers page

DZGoldman:
(shoutout to Winnie). 
Maybe there's a better place for this, but i figure here b/c it should the chain relationships, idk.


TucksonDev (commented):
I think it's a good page to have that graphic, although not sure about its location. @symbolpunk sure has good ideas about where to place it 💪 

Another page to have this graphic could be the [Arbitrum chains page](https://nitro-docs-git-ecosystem-graphic-offchain-labs.vercel.app/for-devs/concepts/public-chains).

PR #654, opened 2023-10-09
Add comments to precompile howto

Jason-W123:


TucksonDev (approved):

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -10,6 +10,12 @@ import PublicPreviewBannerPartial from '../partials/_orbit-public-preview-banner
 
 <PublicPreviewBannerPartial />
 
+:::caution
+
+The tutorial only works when you use eth_call to call the new precompiles, if you want to add non-view/pure methods or call from other contracts, it will break validation, but we will have `how to update the wasm module root tutorial` to fix this soon. If you want to test modifying the STS before we have instructions for wasm module root generation, you can add command `--node.staker.dangerous.without-block-validator` when you start the node.
```
symbolpunk, line 15:
```suggestion
The guidance in this document will work only if you use `eth_call` to call the new precompiles.

This approach will break block validation when adding non-view/pure methods, or when calling from other contracts. Stay tuned - we're working on another document titled _"How to update the Wasm module root"_ to address these use-cases.

If you want to test modifying the State Transition System (STS) before this guidance is available, use the `--node.staker.dangerous.without-block-validator` config flag when you start your node.
```

Calling @TucksonDev @Jason-W123 to keep me honest ^

symbolpunk (changes requested):
🔥

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
TucksonDev: That's great, @symbolpunk , thanks!
My only comment is that the new "How to update the WASM module root" should be a section in this page, instead of a different page, because it's part of the process of adding a new precompile.

symbolpunk (approved):

PR #663, opened 2023-10-11
Fix 403 error on contributions from forks

TucksonDev:
This PR temporarily disables the format checker for contributions coming from forked repositories. For context, contributions from forked repositories that contain formatting errors, throw when trying to write the comment on the PR (the contribution is coming from a different repository which doesn't have rights to access the writing API for this repo).

Until a better solution is found, disabling seems to be the best option to unblock those external contributions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,7 @@ name: Check format
 run-name: Checking format
 
 on:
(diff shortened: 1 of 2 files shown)


symbolpunk (approved):
TY 🙏

PR #698, opened 2023-10-18
Replace the old ArbSepolia block explorer with the Arbiscan version

mahsamoosavi:
Adding https://sepolia.arbiscan.io to the docs 🔥


symbolpunk (approved):
🔥

PR #723, opened 2023-10-25
Hide precompile partial views from search engine

TucksonDev:
When using the search box to search content of the partial views of the precompiles, the actual partial views are part of the results. For example, you can search `setMinimumL2BaseFee`.
This PR fixes that so those partial files do not appear in the results.
This PR also include changes from running the format command.


mahsamoosavi (approved):
🔥

symbolpunk (approved):
🙏

PR #727, opened 2023-10-27
Add customize gas for precompiles

Jason-W123:


symbolpunk:
I'll be starting a review of this shortly, cc @mahsamoosavi @TucksonDev to avoid rework :D

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
       ]
 ```
 
-From the `logs` field, we can see the `topics[0]` is `0xa937..cfa`, which is the event signature of `Hi(address)`. You can verify this using [4 byte directory](https://www.4byte.directory/event-signatures/?bytes_signature=0xa9378d5bd800fae4d5b8d4c6712b2b64e8ecc86fdc831cb51944000fc7c8ecfa) - `topics[1]` is `Your_address`, which is exactly what we defined above!
+## Option 4: Customize gas usage for a specific method
+
+In the above, we introduced how to define a new precompile function. However, if it is simply defined without performing gas collection within the function, it will lead to potential DOS attacks. For this reason, we will continue to introduce how to perform gas collection in precompile. (event does not need to customize the gas cost, because the program will calculate the gas cost of the event when it starts).
```
symbolpunk, line 272:
```suggestion
The above instructions demonstrate how you can define a new precompile function. However, if this new function is simply defined without performing gas collection within the function, your precompile will be vulnerable to Denial-of-Service (DOS) attacks. These attacks exploit the function by flooding it with excessive requests without bearing the computational cost,

To deter this type of attack, you can implement a gas collection mechanism within your precompile. The event itself doesn't need to specify the gas cost; the program will calculate the gas cost when the event's execution is initially triggered.
```

Low-confidence edit

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+
+At the same time, you can also customize a calculation library with lower gas usage than conventional solidity chain contracts based on this document, so that users can use it as a third-party library call with low gas usage.
```
symbolpunk, line 274:
```suggestion
In addition to introducing gas costs where they don't exist, you can also customize gas costs where they're already being incurred. To demonstrate this approach, consider the `GetBalance` method in `ArbInfo.go`:
```

Question: We're using `GetBalance` as an example, and it seems like we're implementing a sort of "override" (in quotes because it's a different method name, but same signature and purpose) that still costs gas, but just customizes the amount of gas that it costs.

If my understanding is correct, the introduction to this section suggests that you want to "incur a gas cost where it isn't yet being incurred" as a mitigation against DOS attacks - but in this example we're "changing the amount of gas where a cost is already being incurred".

My edits are under the assumption that we want to provide guidance for _both_ scenarios - 1) introduce new gas costs to render DOS attacks infeasible, and 2) customize gas costs wherever they're already being incurred. Is this assumption correct? @Jason-W123 @TucksonDev 🙏

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+
+In procompile, we perform gas deduction by calling the `Burn` method of parameter` ctx `in the precompile function, such as the `GetBalance` method in ArbInfo.go:
```
symbolpunk, line 276:
```suggestion
```

(pulled into the preceding paragraph)

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+```
+// GetBalance retrieves an account's balance
+func (con ArbInfo) GetBalance(c ctx, evm mech, account addr) (huge, error) {
+	if err := c.Burn(params.BalanceGasEIP1884); err != nil {
+		return nil, err
+	}
+	return evm.StateDB.GetBalance(account), nil
+}
+```
+The purpose of this method is to obtain the balance of an address. At the same time, according to the definition in [EIP1884](https://eips.ethereum.org/EIPS/eip-1884), the operation code gas usage of obtaining the address balance is 700 gas, so before returning Before the result, the function consumed 700 gas by calling `c.Burn(int64)` (the return value of `params.BalanceGasEIP1884` is 700).
```
symbolpunk, line 287:
```suggestion
The purpose of this method is to retrieve the balance of an address. As defined in [EIP1884](https://eips.ethereum.org/EIPS/eip-1884), the operation code (opcode) for obtaining the address balance has an associated gas cost of 700 gas. The function accounts for this cost by deducting the specified amount of gas, indicated by the protocol constant `BalanceGasEIP1884`, which is set to `700`, through the call to `c.Burn(int64)`.
```

Question: Is this constant set by Geth, or by us? (would like to link to source, if possible)

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+
+Next, we will try to define a function that consumes gas based on the caller address balance as an example. In this step, we will not charge gas according to the specifications in EIP1884, but reduce the gas usage by 400 to 300 gas.
```
symbolpunk, line 289:
```suggestion
To customize the gas cost, let's implement an alternative to `GetBalance`, called `GetBalanceCustom`:
```

Question: Why is the call to `GetBalance` moved into the scope of `c.Burn` - is this a necessary / meaningful relocation?

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+func (con ArbInfo) GetBalanceCustom(c ctx, evm mech, account addr) (huge, error) {
+	gasForBalanceCall := uint64(300)
+	if err := c.Burn(gasForBalanceCall); err != nil {
+		return evm.StateDB.GetBalance(account), err
+	}
+	return balance, nil
+}
+```
+
+How to register a new precompile method please see option1.
```
symbolpunk, line 302:
```suggestion
To register this new precompile method, refer to Option 1 above.
```

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+
+:::info
+
+Note that the instructions provided in [How to run a full node](/node-running/how-tos/running-a-full-node) **will not** work with your Orbit node. See [Command-line options (Orbit)](/launch-orbit-chain/reference/command-line-options) for Orbit-specific CLI flags.
+
+:::
+
+### Send the transaction and get the transaction receipt
+
+In order to obtain the gas used, we can use the `eth_sendRawTransaction` rpc method to test execution on the chain.
```
symbolpunk, line 320:
```suggestion
In order to obtain the gas used, we can use the `eth_sendRawTransaction` RPC method to test execution on the chain.
```

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+First call:
```
symbolpunk, line 321:
```suggestion
First, call:
```

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+```
+cast send 0x0000000000000000000000000000000000000065 "GetBalance()({Any_Address})"
+```
+Secondly call:
```
symbolpunk, line 325:
```suggestion
Then, call:
```

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+```
+cast send 0x0000000000000000000000000000000000000065 "GetBalanceCustom()({Any_Address})"
+```
+
+Afterwards we will obtain the following two results respectively.
```
symbolpunk, line 330:
```suggestion
The two responses will look like this, respectively:
```

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
```diff
@@ -266,7 +267,123 @@ Note the `logs` field within the transaction receipt:
...
+      "logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+      "status":"0x1",
+      "to":"0x0000000000000000000000000000000000000065",
+      "transactionHash":"{Your_txHash}",
+      "transactionIndex":"0x1",
+      "type":"0x2"
+   }
+}
+```
+We can get that the gas used for execution of tx1 is `gasUsed - gasUsedForL1 = 22938`. Similarly, the gas used for tx2 is `22538` (for the gas model, see [how to estimate gas](/devs-how-tos/how-to-estimate-gas)), subtract the two, and the result is `400`, which is exactly what we expected.
```
symbolpunk, line 386:
```suggestion
Here we can see that the gas cost incurred by the execution of the first transaction is `gasUsed - gasUsedForL1 = 22938`. Similarly, the gas cost incurred by the execution of the second transaction is `22538`. If you Subtract the two, the result is `400`, as expected.

To learn more about the gas cost model, see [how to estimate gas](/devs-how-tos/how-to-estimate-gas). 
```

symbolpunk (changes requested):
Some silly questions and nits 😆

arbitrum-docs/launch-orbit-chain/how-tos/customize-precompile.mdx
Jason-W123: Thanks @symbolpunk , I made this to `GetBalance` is because it would be easier to do that. Follow your thought, if we use `SayHi` will better?

symbolpunk: @Jason-W123 I think `GetBalance` is a great example because it represents more of a real-world situation. I'm trying to confirm that we want to demonstrate both scenarios:

1. Create new gas costs
2. Update existing gas costs 

As long as my suggested edits (including the below edits) align with your original intent, I think we're on the same page 🙏

Jason-W123: Yes, we just want to demonstrate how to add gas cost to a precompile method, so I use a existing method as example, so it would be easy for devs to understand

symbolpunk (approved):

PR #737, opened 2023-11-02
Support i18n

Jason-W123:
Add a script to put untranslated files and translated files together at the target dir


symbolpunk:
Thanks so much @Jason-W123 - LGTM. I've added the locale selector to the navbar here, but I've commented it out until we have some docs localized :)

symbolpunk (approved):

PR #761, opened 2023-11-15
Add FAQs as structured data

TucksonDev:
This PR is an attempt to render the different FAQ pages as [Structured Data](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data). This basically improves discoverability of the FAQs in a search engine (or similar) context.

The most important changes in this PR are:
- Adds a new component: FAQStructuredData
- Changes the FAQ pages, so it uses that component, along with the questions stored in a JSON file (explained in next point)
- Changes the update.ts script in the notion-docs folder, so it generates the JSON files for each FAQ section, instead of rendering the questions in the partial views.
- Partial views are no longer used, but the possibility of reusing the different FAQ sections elsewhere is preserved (instead of calling the partial view, you would call the FAQStructuredData component with the JSON file of the questions).
- FAQs are now rendered in HTML instead of Markdown (although only the title was being rendered in Markdown before)

[Preview building FAQs](https://nitro-docs-git-faqs-structured-data-offchain-labs.vercel.app/for-devs/troubleshooting-building)


symbolpunk (approved):
Very cool

TucksonDev:
It looks like anchors are not working properly. I'll mark this as a draft to avoid merging by accident.

TucksonDev:
Anchors are now working 💪 .
Because we are now generating the FAQs dynamically, when accessing the URL of the page with an anchor (something we do 99% of the times we reference this page), the page would not find that anchor because it would not be present on DOM creation.
I've added a small code to scroll to the specified anchor, once the new component has finished rendering.

Now the scrolling effect works just as it did before.

TucksonDev:
@mahsamoosavi , when you have a minute, could you take a look at this in case I've missed something else? 🙏

mahsamoosavi:
> @mahsamoosavi , when you have a minute, could you take a look at this in case I've missed something else? 🙏

This is interesting! Ty @TucksonDev!

mahsamoosavi (approved):

PR #776, opened 2023-11-28
Add more Stylus FAQs

TucksonDev:
This PR adds a few more Stylus FAQs

[Preview](https://nitro-docs-git-update-stylus-faqs-offchain-labs.vercel.app/stylus/troubleshooting-building-stylus)


website/static/building-stylus-faqs.json
```diff
@@ -2,5 +2,10 @@
...
+{"question": "Do Stylus contracts compile down to EVM bytecode like prior other attempts?","answer": "<p>No. Stylus contracts are compiled down to WASM. The user writes a program in Rust / C / C++ which is then compiled down to WebAssembly.</p>\n\n<p></p>\n\n","key": "do-stylus-contracts-compile-down-to-evm-bytecode-like-prior-other-attempts"},
+{"question": "How is a Stylus contract deployed?","answer": "<p>Stylus contracts are deployed on chain as a blob of bytes, just like EVM ones. The only difference is that when the contract is executed, instead of invoking the EVM, we invoke a separate WASM runtime. The way you tell apart the two is by a special EOF-inspired prefix. When a contract code starts with the magic <code>0xEF000000</code> prefix, it's a WASM. </p>\n\n<p></p>\n\n","key": "how-is-a-stylus-contract-deployed"},
```
symbolpunk, line 7:
```suggestion
{"question": "How is a Stylus contract deployed?","answer": "<p>Stylus contracts are deployed on chain as a blob of bytes, just like EVM ones. The only difference is that when the contract is executed, instead of invoking the EVM, we invoke a separate WASM runtime. Note that a special EOF-inspired prefix distinguishes Stylus contracts from traditional EVM contracts: when a contract's bytecode starts with the magic <code>0xEF000000</code> prefix, it's a Stylus WASM. </p>\n\n<p></p>\n\n","key": "how-is-a-stylus-contract-deployed"},
```

nits

symbolpunk (approved):
LGTM! Minor nits noted; we can apply these to Notion separately if preferred 👍

PR #813, opened 2023-12-15
use sepolia for quick start

Jason-W123:
Because many new devs may use this quickstart to deploy contract on arbitrum testnet, we should encourage them to use arb-sepolia instead of arb-goerli. (question: should we use `$ASPL` as a shorthand for "Arbitrum Sepolia testnet $ETH"?)


symbolpunk (approved):
TYSM for this rev - love the usage of `$ASPL` 👍

PR #894, opened 2024-01-22
add configuration parameters

leartulaj:


symbolpunk:
_Marking as `Draft` while we align on possibly refactoring into a reference doc_

symbolpunk (approved):
TYSM!

PR #908, opened 2024-01-26
Update multicall contract on ArbOne and format docs

TucksonDev:
This PR:
- Updates the ArbMulticall contract address for ArbOne (it was not pointing at the right one)
- Format all current docs


symbolpunk (approved):
TYSM

PR #909, opened 2024-01-26
chore: bump Arbitrum SDK version

### LINKED ISSUE #371: Bump node version to 18
Node 18 has been live for more than a year (current LTS): https://github.com/nodejs/node/commits/main/doc/changelogs/CHANGELOG_V18.md

~Node 20 has been live for a few months: https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md#20.0.0~

TucksonDev:
Closes #371 

Once this is merged, we can update the deployment settings on Vercel to Node 18.


symbolpunk:
@TucksonDev probably a silly q, but does Vercel's version need to be bumped before this will deploy? From Vercel logs: 

> error nitro-docs@0.0.0: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.20.2"

TucksonDev:
Oh, right. Vercel complains that we are now enforcing a higher node version, but we still haven't changed their config.
I just reverted that change, so the repo can use versions >= 16 (as it was before). We can then merge this, and change the config in Vercel.

symbolpunk (approved):
TYSM

PR #980, opened 2024-02-14
chore: regenerate precompiles reference + format

TucksonDev:
Since the nitro version was updated, precompiles need to be regenerated to point at the right github resources.

This PR also formats previous unformatted pages.


symbolpunk (approved):
🙏

PR #984, opened 2024-02-15
Clarify arbOS version used when upgrading vs when obtaining it from ArbSys

TucksonDev:
Obtaining the current ArbOS version from ArbSys yields a result that follow a different behavior than when scheduling an ArbOS upgrade.
ArbSys will add 55 to the current version, but you use the actual version (without 55) when scheduling the upgrade.

This PR adds clarification around that behavior.

[Preview](https://arbitrum-docs-git-feat-clarify-arbos-version-offchain-labs.vercel.app/launch-orbit-chain/how-tos/arbos-upgrade)


symbolpunk (approved):
Thank you sir

PR #986, opened 2024-02-15
feat: add precompile addresses to the reference section + small note in the Troubleshooting node page

TucksonDev:
This PR adds all precompile addresses to the precompiles reference page. I find it extremely useful to have the addresses there while navigating through the interfaces or the implementation of the different functions, so I know where the precompile is located.

This PR also adds a small note about removing sensitive information from the `docker run` command when filling up the information in the Troubleshooting: Run a node page

[Preview](https://arbitrum-docs-git-precompiles-reference-an-bb2192-offchain-labs.vercel.app/for-devs/dev-tools-and-resources/precompiles)


symbolpunk (approved):
👍

PR #991, opened 2024-02-15
Change example method of generating BLS keys to docker

a-thomas-22:
Using kubernetes as the base example to generate the BLS keys seems overly complex and not useful to teams not running kubernetes.

This streamlines the process a bit and should reduce confusion on generating the keys.

### DIFF
--- a/arbitrum-docs/node-running/how-tos/data-availability-committee/deploy-a-das.mdx
+++ b/arbitrum-docs/node-running/how-tos/data-availability-committee/deploy-a-das.mdx
@@ -102,7 +102,7 @@ The BLS keypair must be generated using the `datool keygen` utility. Later, it
 
 When running the key generator, we'll specify the `--dir` parameter with the absolute path to the directory inside the volume to store the keys in. 
 
-Here's an example of how to use the `datool keygen` utility inside docker and store the key that will be used by the DAS in the next step.
+Here's an example of how to use the `datool keygen` utility inside Docker and store the key that will be used by the DAS in the next step.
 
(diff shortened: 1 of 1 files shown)


arbitrum-docs/node-running/how-tos/data-availability-committee/deploy-a-das.mdx
```diff
@@ -94,81 +94,22 @@ In terms of hardware requirements, these are the minimum specs that your infrast
...
 
 The BLS keypair must be generated using the `datool keygen` utility. Later, it will be passed to the DAS by file or command line.
 
-When running the key generator, we'll specify the `--dir` parameter with the absolute path to the directory inside the volume to store the keys in. That directory will need to exist before running the tool.
+When running the key generator, we'll specify the `--dir` parameter with the absolute path to the directory inside the volume to store the keys in. 
 
-Here's an example of how to use a k8s deployment to run the `datool keygen` utility and store the key on the volume that we created in the previous step (and that will be used by the DAS in the next step). Note that after this deployment has run once, the deployment can be torn down and deleted:
+Here's an example of how to use the `datool keygen` utility inside docker and store the key that will be used by the DAS in the next step.
```
symbolpunk, line 105:
```suggestion
Here's an example of how to use the `datool keygen` utility inside Docker and store the key that will be used by the DAS in the next step.
```

symbolpunk (approved):
Love it; calling @TucksonDev to review in case there's a reason we oriented this step around kubernetes 🙏

PR #1011, opened 2024-02-22
Update stylus-gentle-introduction.md

yjshi2015:
it means "When mainnet" ?


symbolpunk:
Thanks for the PR - "Wen merge" is a recognized colloquial expression among our readership; I don't think it impedes understanding, so I'm opting to keep as-is.

yjshi2015:
> Thanks for the PR - "Wen merge" is a recognized colloquial expression among our readership; I don't think it impedes understanding, so I'm opting to keep as-is.

That's it, very interesting～

PR #1014, opened 2024-02-23
docs: refactor and update arbos docs for arbos20

leeederek:
- reformats how the docs are structured for arbos information by splitting up the arbos releases into dedicated sub-sections
- updates arbos20 section with github release and audit report
- rename the ethereum l1 rpc provider page to be more explicit that this section covers beacon chain endpoints


mahsamoosavi (approved):
LGTM

symbolpunk:
@leeederek when renaming/moving files in future PRs, we need to handle redirects appropriately. Not a big deal for this doc since it's not a super popular doc, but this will break external links that point to the old doc.

I'd ask that when in doubt, you hold off on merging docs PRs until I have a chance to catch things like this. cc @mahsamoosavi for viz 🙏

PR #1027, opened 2024-02-27
Slugfix - Gentle intro

symbolpunk:
- moves gentle intro to `/welcome/` per new sidebar structure
 - fixes links & adds redirects accordingly
 - cleans up contrib guidance per new sidebar structure
 - adds @anegg0 to `CODEOWNERS`


vercel.json
```diff
@@ -1,5 +1,15 @@
 {
     "redirects": [
+        {
+            "source": "/intro",
+            "destination": "/welcome/arbitrum-gentle-introduction",
+            "permanent": true
```
TucksonDev, line 6: I prefer to set this to `false` always, since we change URLs sometimes and might want to redirect somewhere else in the future.

TucksonDev (approved):
LGTM. Added a small detail to consider, but non-blocking.

vercel.json
symbolpunk: Ty sir - my understanding is that non-permanent redirects are for explicitly _temporary_ changes (eg a/b testing, temporary maintenance, promotions, etc) that indicate 1) we plan to revert and 2) we do not want to pass SEO authority to the new location; does this track with your understanding?

PR #1039, opened 2024-02-28
Sidebar tweak - reference docs across sections

symbolpunk:
- introduces a cross-section doc reference pattern to improve discoverability of:
    - Orbit node-running doc from the _Run an Arbitrum node_ section
    - ArbOS releases docs from the _Run an Orbit chain_ section
    - Data Availability Committees docs from the _Run an Orbit chain_ section
- tested unicode icon rendering on MacOS & Windows
- intent is to continue iterating on UI/UX; this is a stopgap solution

image


hkalodner (approved):
LGTM

PR #1042, opened 2024-02-29
update portal url

Jason-W123:


TucksonDev (approved):
Thanks @Jason-W123 ! 🚀 
(Nice usage of variables btw 💪 )

Issue #928, opened 2024-02-01
Docs update request: /launch-orbit-chain/concepts/anytrust-orbit-chain-keyset-generation


chrisco512:
Source: https://docs.arbitrum.io/launch-orbit-chain/concepts/anytrust-orbit-chain-keyset-generation

The "Inside AnyTrust" link is broken.

anegg0:
This issue has been fixed.

PR #1048, opened 2024-03-01
Fix link 'Inside Anytrust'

anegg0:
Fixed `Inside Anytrust` link which was strangely coded as: `<a data-quicklook-from='inside-anytrust'>Inside Anytrust</a>`
This resolves [that issue](https://github.com/OffchainLabs/arbitrum-docs/issues/928)


anegg0:
It must be redone as I hadn't properly set up my git `user.email` locally.

PR #1049, opened 2024-03-01
Fixed malformed `Inside Anytrust` link

anegg0:
Fixed `Inside Anytrust` link which was strangely coded as: `<a data-quicklook-from='inside-anytrust'>Inside Anytrust</a>`
This resolves [that issue](https://github.com/OffchainLabs/arbitrum-docs/issues/928) .


hkalodner:
@cla-bot check

TucksonDev (approved):
Thanks @anegg0 !

For more context, quicklooks are references that we can add to certain terms to quickly describe them. The description is taken from the Glossary that already exists in the docs.
You can see an example of a quicklook in [Get started](https://docs.arbitrum.io/welcome/get-started), see the first "Arbitrum" which is highlighted and when hovering over it, it gives you a short description loaded from the Glossary.

In any case, your fix is correct since that link should point to "Inside AnyTrust" (and also the quicklook was not correctly defined). We can later add a quicklook to the first AnyTrust mention in this page if needed, but it's not required right now.

TucksonDev (changes requested):
```
[Inside Anytrust](/inside-arbitrum-nitro/inside-arbitrum-nitro.mdx#inside-anytrust)
```

Actually, sorry, I just realized that the link is including the domain, which it shouldn't. Also, it's good to reference the .mdx (or .md) so it can be checked by some internal tooling (for example, if some links are broken). This should be the right one (but test it first, please)

anegg0 (commented):
> ```
[Inside Anytrust](/inside-arbitrum-nitro/inside-arbitrum-nitro.mdx#inside-anytrust)
```

Actually, sorry, I just realized that the link is including the domain, which it shouldn't. Also, it's good to reference the .mdx (or .md) so it can be checked by some internal tooling (for example, if some links are broken). This should be the right one (but test it first, please) 

Thank you for clarifying, @TucksonDev!
I'll keep those pointers in mind (excluding domains and referencing .md/.mdx files, and I even got a chance to use quicklook on other pages.

anegg0:
This was a misunderstanding on my end. 
Apologies to anyone involved!

PR #1051, opened 2024-03-01
docs: add arbos upgrade requirements

leeederek:
adds arbos upgrade requirements


arbitrum-docs/node-running/reference/arbos-software-releases/arbos20.mdx
```diff
@@ -5,9 +5,12 @@ sidebar_position: 2
 author: dlee
 ---
 
-ArbOS 20 Atlas is shipped via Nitro v2.3.0, which is available on Docker hub with the image tag: `offchainlabs/nitro-node:v2.3.0-3e14543`. This release of Nitro is a mandatory upgrade for Arbitrum One and Nova validators.For Arbitrum One and Nova, the ArbOS 20 upgrade requires a governance vote to activate.
+ArbOS 20 Atlas is shipped via Nitro v2.3.1, which is available on Docker hub with the image tag: `offchainlabs/nitro-node:v2.3.1-26fad6f`. This release of Nitro is a mandatory upgrade for Arbitrum One and Nova validators.For Arbitrum One and Nova, the ArbOS 20 upgrade requires a governance vote to activate.
```
TucksonDev, line 8: If v2.3.0 is the first version to support ArbOS 20, shouldn't we mention that version here?
Maybe changing the text to something like "supported since Nitro v2.3.0 ..."

leeederek: Hey good catch, 2.3.0 is indeed the minimum version but we made some critical fixes in 2.3.1 that are important for validators so I decided to just use 2.3.1.

symbolpunk (approved):

symbolpunk:
Approving to unblock, will defer to @TucksonDev's judgment RE how we're communicating the `PATCH` bugfix.

It may make sense to clarify along the lines of his recommendation - _To use ArbOS 20, we recommend upgrading to v2.3.1_ - we can then just link to the relnotes, which presumably clarify the bugfixes. This way we're not implying that the ArbOS 20 upgrade was introduced in a `PATCH` version, and we're explicitly guiding readers to upgrade if they already have 2.3.0.

leeederek:
@symbolpunk @TucksonDev - great flags, all valid points. 

The nuance here is that 2.3.0 is the minimum version for the majority of our audience, but 2.3.1 includes some critical fixes for validators. Validators are a smaller portion of the audience but are key players in securing Arbitrum, so I opted to pin the minimum version to 2.3.1 in our documentation. 

Would definitely welcome advice on how to change the wording. I care about making it very simple for a node operator and validator to upgrade, and that includes simplifying how they consume our docs.

PR #1093, opened 2024-03-18
add a note for cl port

Jason-W123:
Because many users connect beacon url to their execution client's port, we should add a note here to notify them.


arbitrum-docs/node-running/how-tos/running-a-full-node.mdx
```diff
@@ -50,6 +50,7 @@ Even though there are alpha and beta versions of the <a data-quicklook-from='arb
   - It must provide a standard layer 1 node RPC endpoint that you run yourself or from a node provider.
   - Note: this parameter was called `--l1.url` in versions prior to `v2.1.0`
   - Note: the Ethereum consensus layer client Caplin is not compatible with Nitro node software due to an issue with response type formatting.
+  - Note: 8545 is usually the default port for the execution layer. For beacon endpoint port you should connect to the correct one which set on your parent chain consensus client.
```
mahsamoosavi, line 53:
```suggestion
  - Note: 8545 is usually the default port for the execution layer. For the Beacon endpoint port, you should connect to the correct port that is set on your parent chain's consensus client.
```

mahsamoosavi (changes requested):
smol suggestion and then it LGTM 🙏

mahsamoosavi (approved):

PR #1112, opened 2024-03-22
Update stylus-gas.md

chrisco512:
Corrects ink price statement to reflect what would actually occur (price is halved so that EVM gas buys 2x as much).


TucksonDev (approved):

hkalodner:
@cla-bot check

PR #1154, opened 2024-04-04
docs: make explicit the requirements for historical blob data to sync nitro nodes

leeederek:
This PR clarifies some language around what types of node operators need historical blob data to sync up to an Arbitrum chain's latest state.


mahsamoosavi (approved):

symbolpunk (approved):
No blocking feedback.

PR #1179, opened 2024-04-12
Added parameter table to l1-to-l2-messaging page

anegg0:
The parameters listed in this article were rendered as code instead of a list, so I converted this text block to a table.


mahsamoosavi (approved):
Looks way better with that table!

hkalodner (approved):
LGTM

PR #1260, opened 2024-05-15
add Note-on-sequencer-feed to orbit node running

Jason-W123:
Because some orbit node operator follows this docs and said their node doesn't sync to the latest state, so we can add this note to this docs page

### DIFF
--- a/arbitrum-docs/node-running/how-tos/running-an-orbit-node.mdx
+++ b/arbitrum-docs/node-running/how-tos/running-an-orbit-node.mdx
@@ -124,6 +124,24 @@ Or:
   chmod -fR 777 /data/arbitrum
   ```
 
+### Note on Sequencer feed
+
+The Arbitrum node requires a feed to receive real time ordered transactions from the sequencer. If you don't set feed input url, your node will listen to parent chain inbox contract to get ordered transaction which will cause your node to be unable to synchronize the latest state.
+
+To make your node listen to sequencer feed, you need to set those configurations to your sequencer node:
+  
+  ```shell
+  --node.feed.output.enable=true --node.feed.output.addr=<Sequencer feed url> --node.feed.output.port=<Sequencer feed port>
+  ```
+
+Then set the following configurations to your new node:
+  
+  ```shell
+  --node.feed.input.url=<Sequencer feed url>
+  ```
+
+After that, your node can synchronize the latest state from the sequencer feed.
+
 ## Optional parameters
 
 We show here a list of the parameters that are most commonly used when running your Orbit node. You can also use the flag `--help` for a full comprehensive list of the available parameters.


arbitrum-docs/node-running/how-tos/running-an-orbit-node.mdx
```diff
@@ -124,6 +124,24 @@ Or:
   chmod -fR 777 /data/arbitrum
   ```
 
+### Note on Sequencer feed
+
+The Arbitrum node requires a feed to receive real time ordered transactions from the sequencer. If you don't set feed input url, your node will listen to parent chain inbox contract to get ordered transaction which will cause your node to be unable to synchronize the latest state.
+
+To make your node listen to sequencer feed, you need to set those configurations to your sequencer node:
```
TucksonDev, line 131: I would maybe change this a bit, so it's clear that this is something that needs to be configured in the sequencer, not the orbit full node. Otherwise, since this document is mainly targetted at node runners (not chain owners), they might be confused, and think that they need to add these to their nodes.

"In order for a node to read the sequencer feed, the chain's sequencer needs to be configured with the following parameters:"

arbitrum-docs/node-running/how-tos/running-an-orbit-node.mdx
```diff
@@ -124,6 +124,24 @@ Or:
...
+  
+  ```shell
+  --node.feed.output.enable=true --node.feed.output.addr=<Sequencer feed url> --node.feed.output.port=<Sequencer feed port>
+  ```
+
+Then set the following configurations to your new node:
```
TucksonDev, line 137: I would maybe put these before the ones for the sequencer. I'm not sure though, but I'm thinking that this document will be mainly read by users that want to run a full node (and not chain owners who can modify the config of the sequencer) 🤔

arbitrum-docs/node-running/how-tos/running-an-orbit-node.mdx
```diff
@@ -124,6 +124,24 @@ Or:
...
+  --node.feed.input.url=<Sequencer feed url>
+  ```
+
+(Chain owners only) In order for a node to read the sequencer feed, the chain's sequencer needs to be configured with the following parameters:
+  
+  ```shell
+  --node.feed.output.enable=true --node.feed.output.addr=<Sequencer feed url> --node.feed.output.port=<Sequencer feed port>
+  ```
+
+After that, your node can synchronize the latest state from the sequencer feed.
```
TucksonDev, line 143: I think it's clearer if we put this line before the "(Chain owners only)" paragraph, what do you think?

TucksonDev (approved):
LGTM! Added one more comment, but it's now clearer.
Thanks @Jason-W123 !

PR #1261, opened 2024-05-15
updates arbitrum-sdk docs generation 

douglance:
This PR updates the generation of docs for the arbitrum-sdk to use the integrated typedoc plugin for docusaurus as well as the typedoc-markdown plugin in order to output markdown. 

With this system, we no longer need to post-process the `arbitrum-sdk` content. The file that did that post processing has been removed.

This system also allows for us to integrate markdown files from the `arbitrum-sdk` repo itself. A custom Typedoc plugin is included in this PR (`sdkDocsHandler.ts`) that moves files from the `arbitrum-sdk/docs` to the `arbitrum-docs` folder and generates a `sidebar.js` which is imported into the main `sidebar.js`.

In the future, we can use this same system (with minor modifications) for `orbit-sdk` or any other similar TypeScript-based library that we'd like to include in our documentation.

### DIFF
--- a/arbitrum-docs/partials/_troubleshooting-stylus-partial.md
+++ b/arbitrum-docs/partials/_troubleshooting-stylus-partial.md
@@ -158,9 +158,9 @@
 
 <p>
   Since{' '}
-  <a href="https://github.com/OffchainLabs/cargo-stylus/releases/tag/v0.2.1">cargo-stylus v0.2.1</a>
-  , Stylus projects are structured as libraries, so when trying to compile old projects you might
-  get an error <code>no library targets found in package</code>.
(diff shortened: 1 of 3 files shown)


TucksonDev (approved):
LGTM, but I'd like @symbolpunk to take a quick look at it too.
A few minor things:
- Can you run `yarn format` ?
- Would adding [these](https://github.com/OffchainLabs/arbitrum-docs/blob/master/website/typedoc.js#L53-L57) solve the issue with the right sidebar? (not sure if these parameters are outdated)

Outside this PR, the Product team might want to take a look at the current structure, specially the index file, to probably format it in a different way (with a selling introduction, a clearer TOC, ...)

Thanks very much, sir! 🚀

PR #1279, opened 2024-05-21
updates to sdk v4 docs

douglance:
Updates system for generating sdk docs to copy over docs saved in that repo, then generates more docs and puts them into a `Reference` folder. 

Also, does some minimal updates to some of the older docs pages for v4. We could rewrite those to be more parent-child focused in a follow up PR.


anegg0 (approved):
lgtm, thanks for helping to integrate Arbitrum-orbit-sdk

PR #1286, opened 2024-05-23
Update build nitro locally and update nitro version

Jason-W123:
1. tell ppl to download foundry.
2. Use golang version v1.20 because v1.21 is not compatible with ubuntu 22.04:  https://github.com/OffchainLabs/nitro/issues/2311
3. tell people to add flag` --validation.wasm.allowed-wasm-module-roots={}` if they build v2.3.4
4. update nitro version

### DIFF
--- a/arbitrum-docs/run-arbitrum-node/nitro/01-build-nitro-locally.md
+++ b/arbitrum-docs/run-arbitrum-node/nitro/01-build-nitro-locally.md
@@ -199,7 +199,7 @@ To run your node using the generated binaries, use the following command from th
 
 #### WASM module root error (v2.3.4 or later)
 
-If your node shutdown due to this error
+Since v2.3.4, the State Transition Function (STF) contains code that is not yet activated on the current mainnet and testnet chains. Because of that, you might receive the following error when connecting your built node to those chains:
 
 ```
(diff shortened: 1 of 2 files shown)


arbitrum-docs/run-arbitrum-node/nitro/01-build-nitro-locally.md
```diff
@@ -6,7 +6,7 @@ sidebar_position: 7
 content_type: how-to
 ---
 
-import PublicPreviewBannerPartial from '../../partials/_public-preview-banner-partial.md';
+import PublicPreviewBannerPartial from '../../partials/\_public-preview-banner-partial.md';
```
TucksonDev, line 9: I think that `\` excaping character is not needed here. Can you verify that your formatter is using the config from this repo (and not something else?)

arbitrum-docs/run-arbitrum-node/nitro/01-build-nitro-locally.md
```diff
@@ -189,3 +196,17 @@ To run your node using the generated binaries, use the following command from th
 ```bash
 ./target/bin/nitro <node parameters>
 ```
+
+#### WASM module root error
```
TucksonDev, line 200: I think this section could be transformed into an admonition ("info", for example).
And we should let the reader know this is a temporary issue with v2.3.4 only, as we want to avoid users modifying the WASM Module root used by default, imo.

What do you think?

TucksonDev (approved):

PR #1422, opened 2024-07-07
Create deno.yml

Essenceone1:
Essenceone1 entertainment 
DIMEJI gas and co


hkalodner:
Thanks for the contribution, but we don't incorporate external linters into CI

PR #1428, opened 2024-07-10
Add Nova to nodes provided by Alchemy + format

TucksonDev:
[Preview](https://arbitrum-docs-git-add-nova-to-alchemy-node-offers-offchain-labs.vercel.app/build-decentralized-apps/reference/node-providers#third-party-rpc-providers)


arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.md
```diff
@@ -97,15 +97,18 @@ Given that participation in BoLD is permissionless, it is recommended that the s
...
...n finally to the level of a single step of execution. These values were carefully calculated to optimize for the resource ratio and gas costs in the event of an attack, as explained in the [Economics of Disputes in Arbitrum BoLD](https://github.com/OffchainLabs/BoLD/blob/main/docs/research-specs/Economics.pdf) and the [BoLD whitepaper](https://arxiv.org/abs/2404.10491). This effectively means that an entity that has already put up a bond to propose an assertion does not need to put up a separate assertion bond to challenge an invalid state assertion that they observe. To be explicitly clear, a validator would still require 555 ETH and 79 ETH for on-going challenges. These additional challenge bond amounts are needed to participate in the interactive dispute game (back and forth) and narrows down the disagreement to a single step of execution that is then proven on Ethereum. The 555 ETH and 79 ETH challenge bonds can be put together via a trustless bonding pool, and does not all have to be put up by the validator that opened the challenge. These bonds can be refunded at the end of a challenge and can also alternatively be put together by the community using a trustless bonding pool.
```
anegg0, line 102:
```suggestion
- Challenge bonds, per level: 555/79 ETH (UPDATED) - required from validators to open challenges against an assertion observed on Ethereum for each level. Note that “level” corresponds to the level of granularity at which the interactive dissection game gets played over, starting at the block level, moving on to a range of `WASM` execution steps, and then finally to the level of a single execution step. These values were carefully calculated to optimize for the resource ratio and gas costs in the event of an attack, as explained in the [Economics of Disputes in Arbitrum BoLD](https://github.com/OffchainLabs/BoLD/blob/main/docs/research-specs/Economics.pdf) and the [BoLD whitepaper](https://arxiv.org/abs/2404.10491). This effectively means that an entity that has already posted a bond to propose an assertion does not need to submit a separate assertion bond to challenge an invalid state assertion. To be more specific, a validator would still require 555 `ETH` and 79 `ETH` for ongoing challenges. These additional challenge bond amounts are needed to participate in the interactive dispute game (back and forth) and narrow down the disagreement to a single step of execution that is then proven on Ethereum. The 555 `ETH` and 79 `ETH` challenge bonds can be put together via a trustless bonding pool and do not all have to be posted by the validator who opened the challenge. These bonds can be refunded at the end of a challenge and can alternatively be put together by the community using a trustless bonding pool.
```

anegg0 (approved):

PR #1514, opened 2024-07-29
Removal stale placeholder DAC configuration

anegg0:
The removed article is unlikely to be added since readers can find DAC config info at:
https://docs.arbitrum.io/launch-orbit-chain/how-tos/orbit-sdk-deploying-anytrust-chain#4-setting-valid-keyset-on-parent-chain


hkalodner (approved):
LGTM

PR #1525, opened 2024-08-01
Drop code owners file for faster iterations

anegg0:
Although best practice, restricting approval reviews to a subset of individuals throttles the pace of publications.
As an experiment and with Harry's approval, the list will be deleted.


hkalodner (approved):
LGTM

PR #1536, opened 2024-08-04
Rpc endpoints conversion to partials

anegg0:
This PR is preliminary work to enable the creation of a main "chain info" page that will include most references to existing RPC providers, faucets, and contract addresses.

The changes break down most RPC endpoints and testnet info into single sources of truth called `partials`.
Partials can be re-used across the docs while keeping their integrity: they will update themselves just like imported modules, limiting stale or plain misleading refs.

Note:
-   These changes won't affect the UX; only the internal docs' structure.
-   References to the deprecated stylus-testnet have been removed to avoid confusion.


mahsamoosavi (approved):
LGTM

PR #1546, opened 2024-08-07
fix: rephrase Stylus performance-related assertions

anegg0:
Rephrase performance-increase claims until we can share benchmarks.


hkalodner (approved):
LGTM

PR #1554, opened 2024-08-12
updating call in stylus

GreatSoshiant:
In the new Stylus Rust SDK we need to use Call::new_in() instead of Call:new(). In this PR we've corrected these on the doc


mahsamoosavi (approved):
LGTM

PR #1577, opened 2024-08-20
Fix precompiles reference generator

TucksonDev:
This PR fixes the `yarn generate_precompiles_ref_tables` by upgrading the Typescript version to 5.1 ([as suggested in the documentation](https://docusaurus.io/docs/3.3.2/typescript-support)) and tweaking the configuration used when running standalone scripts like that one.


anegg0 (approved):
LGTM

PR #1580, opened 2024-08-21
Stylus sidebar restructure

TucksonDev:
This PR attempts to restructure the sidebar menu for the Stylus section. It includes the following:
- Reorder the menu entries for a shorter sidebar and meaningful subsections to increase discoverability
- Adds the "Caching strategy" page to the "Gas and Ink" subsection, and adapts the Overview to mention it
- Modifies the "Reduce the size of WASM binaries" to remove outdated information and adapt the language to the new "CLI tools" subsection

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -437,19 +437,7 @@ const sidebars = {
         },
         {
           type: 'doc',
-          id: 'stylus/concepts/stylus-cache-manager',
-          label: 'Stylus caching strategy',
-        },
-        {
-          type: 'html',
-          value:
-            '<a class="menu__link menu__list-item" href="/run-arbitrum-node/run-local-dev-node">Run a Stylus dev node<span class="other-section-icon">↑</span></a>',
-          // q: why use an anchor html tag here?/node-running/how-tos/running-an-stylus-node
-          // a: see note at end of file
-        },
-        {
-          type: 'doc',
-          label: 'Testnets',
+          label: 'Testnet',
           id: 'stylus/reference/testnet-information',
         },
         {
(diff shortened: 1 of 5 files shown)


website/sidebars.js
```diff
@@ -437,30 +437,34 @@ const sidebars = {
...
+            {
+              type: 'link',
+              label: 'Rust crate docs',
+              href: 'https://docs.rs/stylus-sdk/latest/stylus_sdk/index.html',
+            },
+          ],
+        },
+        {
+          type: 'category',
+          label: 'Gas, ink and caching',
```
mahsamoosavi, line 467: Should we replace ink with Ink for the sake of consistency?

mahsamoosavi (changes requested):
A tiny comment, LGTM otherwise!

website/sidebars.js
TucksonDev: I see that I haven't used that in other places though. I'll update them.

TucksonDev: Updated, thanks for the flag!

anegg0 (approved):
Looking great to me. The content will evolve within the next two weeks, but I find this new structure much more straightforward.

website/sidebars.js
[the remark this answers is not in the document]
anegg0: Are you OK with keeping ink low-cased, @mahsamoosavi?
I think it's better low-cased.
Also, the concept of `ink` is on its way to being deprecated.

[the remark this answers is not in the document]
mahsamoosavi: Yes let's keep it low-cased.

mahsamoosavi (approved):

PR #1581, opened 2024-08-21
Add tip to fix nitro compilation warnings on MacOS

gligneul:


eljobe (approved):
I'm happy with this change. But, I also wonder if we could detect the OS and just set these environment variables in the Makefile instead of having each developer remember it themselves?

PR #1592, opened 2024-08-26
Add note about potential changes on custom gas tokens

TucksonDev:
Preview

### DIFF
--- a/arbitrum-docs/launch-orbit-chain/how-tos/use-a-custom-gas-token.mdx
+++ b/arbitrum-docs/launch-orbit-chain/how-tos/use-a-custom-gas-token.mdx
@@ -25,6 +25,12 @@ There are other important considerations to keep in mind when deciding to use a 
 - The token must only be able to set allowance via a call to the token address itself.
 - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert.
 
+:::info
+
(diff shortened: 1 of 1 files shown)


arbitrum-docs/launch-orbit-chain/how-tos/use-a-custom-gas-token.mdx
```diff
@@ -25,6 +25,12 @@ There are other important considerations to keep in mind when deciding to use a
 - The token must only be able to set allowance via a call to the token address itself.
 - The token must not have a callback on transfer, and more generally a user must not be able to make a transfer to themselves revert.
 
+:::info
+
+While certain restrictions are in place today (see above), some of them may be removed in upcoming releases of the custom gas token feature. Please reach out to the Offchain Labs team with any questions or requests for additions to the custom gas token feature.
```
anegg0, line 30: Would it make sense to add a link to the OCL channel they can reach out to?

anegg0 (changes requested):
Does it make sense to indicate what OCL channel the reader can reach out to?

arbitrum-docs/launch-orbit-chain/how-tos/use-a-custom-gas-token.mdx
TucksonDev: This is intended for RaaSes, which usually have a comms channels with us. I'm a bit hesitant of adding Discord explicitly here, since I'd want to avoid readers using discord if they already have a different comms channel available. 🤔

anegg0 (approved):
LGTM

PR #1616, opened 2024-08-31
update docs to stylus sdk v6

GreatSoshiant:
Deleted the paragraph about [view], [pure] and [write] macros which are not existing anymore.


TucksonDev (changes requested):
Hey @GreatSoshiant !
Copying here the message I sent on Slack:

The text of that section has been updated as per [the latest changes](https://github.com/OffchainLabs/stylus-sdk-rs/blob/main/stylus-proc/src/lib.rs#L428-L434), so it is valid. It explains when methods are considered view , pure or "write".
But I didn't change the title, and that's what's confusing :sweat_smile:. Can you change the title to "pure, view and write functions" or something like that? (without the macro syntax).

Thanks for the flag! :D

TucksonDev (approved):
LGTM

PR #1637, opened 2024-09-03
Replay tx macos warning

anegg0:
This PR add a warning for MacOS users: the `replay` function won't work in the current version.


TucksonDev (approved):
LGTM

PR #1707, opened 2024-09-25
Format pages

TucksonDev:
This PR just format all pages.

Should we perform a format check on CI and block merging if pages are not well formatted?


anegg0 (approved):
LGTM, thank you, @TucksonDev!

PR #1716, opened 2024-09-30
docs: update docs for arbos32

leeederek:
- updates all links and mentions of arbos31 to arbos32
- updates our guidance to point folks to directly go from ArbOS20 to ArbOS32 and skip arbos30 and arbos31
- accompanied by: https://github.com/OffchainLabs/orbit-actions/pull/27 as well as https://github.com/ArbitrumFoundation/docs/pull/864

Fully acknowledge that 3.2.1 is only required for validators and node operators can stick to 3.2.0, but I've opted to exclude this distinction and ask that all folks upgrade to arbos32.


website/sidebars.js
```diff
@@ -663,8 +663,8 @@ const sidebars = {
             },
             {
               type: 'doc',
-              id: 'run-arbitrum-node/arbos-releases/arbos31',
-              label: 'ArbOS 31 Bianca',
+              id: 'run-arbitrum-node/arbos-releases/arbos32',
```
TucksonDev, line 666: Can you create a redirection in `vercel.json` file? That way, if someone is still referencing the `arbos31` URL, the portal will redirect the user to `arbos32`.

Let me know if you need help creating the redirection 🙏

TucksonDev (changes requested):
LGTM, but added one change. Since links to the `arbos31` page might have been posted externally, we should create a redirection to the renamed file.

website/sidebars.js
leeederek: thanks! done

TucksonDev (approved):
LGTM!

PR #1718, opened 2024-10-01
Fix cast commands in stylus quickstart

SeanZoR:
**Problem:**
1. The forward slash at the end of the `cast call` command breaks the command.
2. Missing backslash at the end of the `cast send`.

**Fix:**
Added/Replaced with backslash.


anegg0 (approved):
LGTM, thanks much, @SeanZoR!

hkalodner:
@cla-bot check

PR #1733, opened 2024-10-09
Stylus quickstart vending machine

anegg0:
This PR adds Mahsa's `vending machine` example to the quickstart guide.


mahsamoosavi (commented):
It would be good to mention at the start that this guide covers both a basic Counter contract and a bit more advanced Vending Machine contract. That way people know they’ll start simple and then move on to another complex example.

PR #1757, opened 2024-10-18
Specify toolchain for wasm target

joshuacolvin0:
`rustup target add` should be run with `--toolchain` option to apply globally so that it doesn't have to be run in every single project directory.


anegg0 (approved):
LGTM, thanks, @joshuacolvin0!

PR #1793, opened 2024-11-05
chore: update glossary

hkalodner:


anegg0 (approved):
Thanks for the additional clean up. LGTM

PR #1906, opened 2024-12-13
Create check-undocumented-issues.yml

anegg0:
This PR's goal is to set up a GitHub Action that will automatically close issues created without any additional content from the author. 
The action will check if the issue is a documentation update request and if it only contains the template content. If so, it will close the issue with a comment guiding the author on providing more detailed information for documentation updates.


hkalodner (approved):
LGTM

PR #1954, opened 2025-01-06
Add stringStorage setter and getter tu stylus-sdk-guide

Jason-W123:
Because stringStorage use different method to set and get its storage value, we need to mention it on docs.


anegg0 (approved):
lgtm

PR #1975, opened 2025-01-17
docs: updating-challenge-manager-diagrams

pete-vielhaber:
- updating diagrams for Interactive Fraud Proofs page


anegg0 (changes requested):
LGTM, the only suggestion is to underline function names. Currently, they are difficult to identify

anegg0 (approved):
lgtm

PR #1983, opened 2025-01-21
Update Orbit fee management page

TucksonDev:
This PR updates the "How to manage the fee parameters of your Orbit chain" page. It updates the old contents (since some of the instructions are now also supported in the Orbit SDK), and adds a new section to configure the current fee parameters.

[Preview](https://arbitrum-docs-git-update-fee-management-page-offchain-labs.vercel.app/launch-orbit-chain/how-tos/manage-fee-collectors)


pete-vielhaber (approved):
Some notes are called out but not as admonitions; however, some are back-to-back with other admonitions, will circle back later to determine if we need to make a change with stacked admonitions.

PR #2001, opened 2025-01-24
adds update variable system to write vars to mdx

douglance:
- adds `update-variable-references.ts` file which can be run to update markdown files when globalVars.js changes
- swaps to using a function in `markdown-preprocessor.js` that will swap out variables in markdown content before AST parsing happens 
- swaps to using `@@` instead of `@` for delimiting the bounds of our injected variables


website/src/remark/variable-injector.js
```diff
@@ -42,17 +42,9 @@ const plugin = (options) => {
...
-      //    @var@
-      //    @ var @
-      //    @my-var @
-      //    @ my-4th_var@
-      value = value.replace(/\@\s*([a-zA-Z0-9_-]+)\s*\@/g, (match, varName) => {
-        return options.replacements[varName] || match;
-      });
+      // Format: @variableName=value@
+      // This extracts just the value part from the variable reference
+      value = value.replace(/\@\s*[a-zA-Z0-9_-]+=([^@]+)\s*\@/g, (match, value) => value.trim());
```
TucksonDev, line 47: It looks like we have an issue with values that contain the ':' character (for example, [here](https://arbitrum-docs-git-dl-var-inject-offchain-labs.vercel.app/run-arbitrum-node/run-full-node#prerequisites)).

After some debugging, it looks like this only happens with "code" tags that contain a ':' character. In those cases, the code AST node creates 3 different text nodes underneath it. In the example linked above it does:
code
   text
       '@latestNitroNodeImage=offchainlabs/nitro-node'
   text
       ':v3'
   text
       '.4.0-d896e9c@'
       
so it doesn't match the pattern.

There's probably a way to merge these 3 text nodes, but I'm not sure how to generalize the logic so it doesn't mess findings in other pages.

website/tsconfig.json
```diff
@@ -3,8 +3,10 @@
   "extends": "@tsconfig/docusaurus/tsconfig.json",
   "compilerOptions": {
     "baseUrl": ".",
+    "esModuleInterop": true,
```
TucksonDev, line 6: Do we need these changes here? Just wondering if this might impact other processes.

TucksonDev (changes requested):
Thanks @douglance !
I'm blocking until we figure out a way to fix the issue with the "code" AST nodes and the ":" character.

website/src/scripts/markdown-preprocessor.js
```diff
@@ -0,0 +1,12 @@
+const globalVars = require('../resources/globalVars.js');
+
+function preprocessContent({ filePath, fileContent }) {
+  if (!filePath.endsWith('.mdx')) return fileContent;
+
+  return fileContent.replace(
+    /@@\s*([a-zA-Z0-9_-]+)=([^@]+)@@/g,
+    (match, varName, value) => globalVars[varName] || value.trim(),
```
TucksonDev, line 8: We should add a comment here (and maybe in the `update-variable-references` too, but at least here), explaining that pages that are not modified are cached by Vercel and not updated, and that's why we follow this logic that seems like "duplication" (in the general case, we won't use the value that is in the content of the page, but the value that is in `globalVars`).

TucksonDev (approved):
Thanks @douglance !
Approving to unblock, but please add the comment I mention in my latest message 🙏

website/src/remark/variable-injector.js
TucksonDev: Solved by changing the way variables are replaced

PR #2020, opened 2025-01-28
adds husky precommit hook for running precompiles

douglance:
adds precommit hook that checks if `globalVars.js` has changes, and if it does, runs the precompile script for you automatically. 

it doesn't prevent your commit from going through, so it shouldn't mess with your workflow too much. it also has a log message to explain how to skip this if it's disruptive.


TucksonDev (approved):
LGTM thanks!

pete-vielhaber (approved):

PR #2027, opened 2025-01-29
Add vanilla admonitions

anegg0:
This PR proposes more discreet admonitions.
Based on contributors' feedback, the default Docusaurus admonitions are too prominent and intrusive. Unfortunatelyf, they cannot be modified (swizzled) safely.
An example (vanilla admonition test page) is in the "welcome" section.


pete-vielhaber (approved):
I believe the Vanilla landing page should be removed before going live. Looks great, thanks for the work on this!

PR #2034, opened 2025-02-01
Content structure generator

anegg0:
This PR aims to better assess the site's content structure by adding a script that generates a representation of `sidebars.js` in a reading-friendly format (contentStructure.txt).

The usage guide is in the script itself (`contentStructureGenerator.js`).


pete-vielhaber (approved):
See if we can add to `yarn format` so we can keep an up-to-date output

PR #2047, opened 2025-02-05
docs: haw-bridging-parent-to-child-messaging

pete-vielhaber:


arbitrum-docs/how-arbitrum-works/06-optimistic-rollup.mdx
```diff
@@ -43,13 +55,13 @@ The alternative to interactive proving would be to have a rollup block contain a
...
+**Higher per-tx gas limit:** Interactive proving can escape from Ethereum's tight per-transaction gas limit. The gas limit isn't infinite, for obvious reasons, but it can be larger than on Ethereum. As far as Ethereum is concerned, the only downside of a gas-heavy Arbitrum transaction is that it may require an interactive <a data-quicklook-from="fraud-proof">fraud proof</a> with slightly more steps (and only if a fraud proof is actually needed; i.e., in the event of a dispute). By contrast, rerexecution must impose a _lower_ gas limit than Ethereum, because it must be possible to emulate execution of the transaction (which is more expensive than executing it directly) within a single Ethereum transaction.
```
anegg0, line 62:
```suggestion
**Higher per-tx gas limit:** Interactive proving can escape Ethereum's tight per-transaction gas limit. The gas limit isn't infinite, for obvious reasons, but it can be larger than on Ethereum. As far as Ethereum is concerned, the only downside of a gas-heavy Arbitrum transaction is that it may require an interactive <a data-quicklook-from="fraud-proof">fraud proof</a> with slightly more steps (and only if a fraud proof is actually needed; i.e., in the event of a dispute). By contrast, re-execution must impose a _lower_ gas limit than Ethereum, because it must be possible to emulate the execution of the transaction (which is more expensive than executing it directly) within a single Ethereum transaction.
```

arbitrum-docs/how-arbitrum-works/06-optimistic-rollup.mdx
```diff
@@ -43,13 +55,13 @@ The alternative to interactive proving would be to have a rollup block contain a
...
 
-**More implementation flexibility:** Interactive proving allows more flexibility in implementation. All that is necessary is the ability to verify a one-step proof on Ethereum. By contrast, reexecution approaches are tethered to limitations of the EVM.
+**More implementation flexibility:** Interactive proving allows more flexibility in implementation. All that is necessary is the ability to verify a one-step proof on Ethereum. By contrast, rerexecution approaches are tethered to limitations of the EVM.
```
anegg0, line 64:
```suggestion
**More implementation flexibility:** Interactive proving allows more flexibility in implementation. All that is necessary is the ability to verify a one-step proof on Ethereum. By contrast, re-execution approaches are tethered to the limitations of the EVM.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 author: pete-vielhaber
-sme: TucksonDev
-user_story: As a current or prospective Arbitrum user, I need to learn more about messaging between L1 and L2 within Arbitrum.
+sme: Mehdi
+user_story: As a current or prospective Arbitrum user, I need to learn more about messaging between a parent and child chain within Arbitrum.
 content_type: get-started
 ---
 
-import { AddressAliasHelper } from '@site/src/components/AddressAliasHelper';
+In the [Bypassing the Sequencer] section, we introduced an alternative way for users to submit transactions to a child chain by going through the parent chain's delayed inbox contract instead of sending them directly to the Sequencer. This approach is one example of a parent to child messaging path. More broadly, parent to child chain messaging covers all ways to:
```
anegg0, line 10:
```suggestion
In the [Bypassing the Sequencer] section, we introduced an alternative way for users to submit transactions to a child chain by going through the parent chain's delayed inbox contract instead of sending them directly to the Sequencer. This approach is one example of a parent-to-child messaging path. More broadly, parent-to-child chain messaging covers all ways to:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-In the [Transaction Lifecycle](/how-arbitrum-works/02-transaction-lifecycle.mdx) section, we covered how users interact with L2 contracts. They submit transactions by putting messages into the chain’s inbox or having a full node <a data-quicklook-from="sequencer">Sequencer</a> or aggregator do so on their behalf.
+- Submit child chain bound transactions from a parent chain
+- Deposit ETH or native tokens from the parent chain to the child chain
+- Send arbitrary data or instuctions from the parent chain to the child chain
```
anegg0, line 14:
```suggestion
- Submit child chain bound transactions from a parent chain
- Deposit ETH or native tokens from the parent chain to the child chain
- Send arbitrary data or instructions from the parent chain to the child chain
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-L1 and L2 chains run asynchronously from each other, so it is not possible to make a cross-chain call that produces a result within the same <a data-quicklook-from="transaction">Transaction</a> as the caller. Instead, cross-chain calls must be asynchronous, meaning that the caller submits the call at some point in time, and the call runs later. Consequently, a cross-chain contract-to-contract call can never produce a result available to the calling contract (except for acknowledgment of a successful call submitted for later execution).
+We generally categorize these parent to child chain messaging methods as follows:
```
anegg0, line 16:
```suggestion
We generally categorize these parent-to-child chain messaging methods as follows:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+- **ETH Bridging**: For Arbitrum chains that use ETH as their gas token, users can deposit ETH onto a child chain via the delayed inbox
+- **Custom gas token bridging**: For Arbitrum chains that use a custom gas token, users can deposit that chain's native token to the child chain using the same mechanism
 
-Retryable tickets are <a data-quicklook-from="arbitrum">Arbitrum</a>'s canonical method for creating L1 to L2 messages, i.e., L1 transactions that initiate a message to be executed on L2. A retryable can be submitted for a fixed cost (dependent only on its calldata size) paid at L1; its _submission_ on L1 is separable / asynchronous with its _execution_ on L2. Retryables provide atomicity between the cross chain operations; if the L1 transaction to request submission succeeds (i.e. does not revert) then the execution of the Retryable on L2 has a strong guarantee to ultimately succeed as well.
+2.  **Transacting via the delayed inbox** as described in the [Bypassing the Sequencer](/how-arbitrum-works/02-transaction-lifecycle.mdx#bypassing-the-sequencer) section, this method, allows users to send transaction through the parent chain. It includes two sub-types of messages:
```
anegg0, line 23:
```suggestion
2.  **Transacting via the delayed inbox** as described in the [Bypassing the Sequencer](/how-arbitrum-works/02-transaction-lifecycle.mdx#bypassing-the-sequencer) section, this method, allows users to send transactions through the parent chain. It includes two sub-types of messages:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-## Retryable Tickets Lifecycle
+- **Unsigned messages**: General arbitrary data or function calls
+- **Signed messages**: Messages that include a signature, enabling certain authenticated actions
 
-Here we walk through the different stages of the lifecycle of a <a data-quicklook-from="retryable-ticket">Retryable Ticket</a>; (1) submission, (2) auto-redemption, and (3) manual redemption.
+3.  **Retryable tickets** are Arbitrum's canonical mechanism for creating parent to child messages–transactions initiated on the parent chain tha trigger execution on the child chain. This method contains the following functionality:
```
anegg0, line 28:
```suggestion
3.  **Retryable tickets** are Arbitrum's canonical mechanism for creating parent-to-child messages–transactions initiated on the parent chain that trigger execution on the child chain. This method contains the following functionality:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-### Submission
+- **General retryable messaging**: For sending arbitrary data or calls from parent to child chain
+- **Customized feature messaging** (e.g., token bridging): Leveraging retryable tickets (and other messaging constructs) for specialized actions, such as bridging tokens from a parent to a child chain
 
-1. Creating a retryable ticket is initiated with a call (direct or internal) to the `createRetryableTicket` function of the [`inbox` contract](https://github.com/OffchainLabs/nitro-contracts/blob/67127e2c2fd0943d9d87a05915d77b1f220906aa/src/bridge/Inbox.sol). A ticket is guaranteed to be created if this call succeeds. Here, we describe parameters that need to be carefully set. Note that, this function forces the sender to provide a _reasonable_ amount of funds (at least enough to submitting, and _attempting_ to executing the ticket), but that doesn't guarantee a successful auto-redemption.
+In this section, we will explore each of these categories in detail and explain how they work. The diagram below illustrates the various paths available for parent to child chain communication asset transfers.
```
anegg0, line 33:
```suggestion
In this section, we will explore each of these categories in detail and explain how they work. The diagram below illustrates the various paths available for parent-to-child chain communication asset transfers.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-### Automatic Redemption
+### Depositing ETH
 
-4. It is very important to note that the submission of a ticket on L1 is separable / asynchronous from its execution on L2, i.e., a successful L1 ticket creation does not guarantee a successful redemption. Once the ticket is successfully created, the two following conditions are checked: (1) if the user's L2 balance is greater than (or equal to) `maxFeePerGas * gasLimit` **and** (2) if the `maxFeePerGas` (provided by the user in the ticket submission process) is greater than (or equal to) the `l2Basefee`. If these conditions are both met, ticket's submission is followed by an attempt to execute it on L2 (i.e., an **auto-redeem** using the supplied gas, as if the `redeem` method of the [ArbRetryableTx](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx) precompile had been called). Depending on how much gas the sender has provided in step 1, ticket's redemption can either (1) immediately succeed or (2) fail. We explain both situations here:
+A special message type exists for simple ETH deposits from a parent to child chain. You can deposit ETH by calling the `Inbox` contract's `depositEth` method, for example:
```
anegg0, line 45:
```suggestion
A special message type exists for simple ETH deposits from a parent-to-child chain. You can deposit ETH by calling the `Inbox` contract's `depositEth` method, for example:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+While `depositEth` is often the simplest path, you can also use _retryable tickets_ to deposit ETH. This may be preferable if you need additional flexibility–for example, specifying an alternative destination address or triggering a fallback function a child chain.
+
+#### How deposits work
+
+When you call `Inbox.depositEth`, the ETH is sent to the bridge contract on the parent chain. This contract then "credits" the deposited amount to the specified address on the child chain. As far as the parent chain is concerned, the deposited ETH remains held by Arbitrum's bridge contract on your behalf.
+
+A diagram illustrating this deposit process can be found below:
+
+- **If the parent chain caller is an EOA** (Externally Owned Account): The deposited ETH will appear in the same EOA address on the child chain.
+- **If the parent chain caller is a contract**: The ETH will be deposited to the contract's **aliased** address on the child chain. Aliasing is described below.
```
anegg0, line 66:
```suggestion
:::note
Depositing ETH directly via `depositEth` to a contract on a child chain **will not** invoke that contract's fallback function.
:::

#### Using retryable tickets instead

While `depositEth` is often the simplest path, you can also use _retryable tickets_ to deposit ETH. This may be preferable if you need additional flexibility–for example, specifying an alternative destination address or triggering a fallback function in a child chain.

#### How deposits work

When you call `Inbox.depositEth`, the ETH is sent to the bridge contract on the parent chain. This contract then "credits" the deposited amount to the specified address on the child chain. As far as the parent chain is concerned, the deposited ETH remains held by Arbitrum's bridge contract on your behalf.

A diagram illustrating this deposit process can be found below:

- **If the parent chain caller is an EOA** (Externally Owned Account): The deposited ETH will appear in the same EOA address on the child chain.
- **If the parent chain caller is a contract**: The ETH will be deposited to the contract's **aliased** address on the child chain. Aliasing is described below.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+All unsigned messages submitted through the delayed inbox have their sender addresses "aliased" when executed on the child chain. Instead of returning the parent chain sender's address as `msg.sender`, the child chain sees the "child alias" of that address. Formally, the child alias is calculated as:
+
+```solidity
+Child_Alias = Parent_Contract_Address + 0x1111000000000000000000000000000000001111
+```
+
+#### Why aliasing?
+
+Address aliasing in Arbitrum is a security measure that prevents cross-chain exploits. Without it, a malicious actor could impersonate a contract on a child chain simply by sending a message from that contract's parent chain address. By introducing an offset, Arbitrum ensures the child chain contracts can distinguish between calls from parent chain contracts and calls from child chain native addresses.
```
anegg0, line 80:
```suggestion
Address aliasing in Arbitrum is a security measure that prevents cross-chain exploits. Without it, a malicious actor could simply impersonate a contract on a child chain by sending a message from that contract's parent chain address. By introducing an offset, Arbitrum ensures the child chain contracts can distinguish between calls from parent chain contracts and calls from child chain native addresses.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+Arbitrum provides a _delayed inbox_ contract on the parent chain that can deliver arbitrary messages to the child chain. This functionality is important for two reasons:
+
+1. **General crosschain messaging**: It allows parent chain EOAs or parent chain contracts to send messages or transactions to the child chain. This is critical for bridging assets (other than the chain's native token) and performing cross-chain operations.
+2. **Censorship resistance**: It ensures the Arbitrum chain remains censorship resistant, even if the Sequencer misbehaves or excludes certain transactions. See [Bypassing the Sequencer](/how-arbitrum-works/02-transaction-lifecycle.mdx#bypassing-the-sequencer) for more details.
+
+Users can send child chain transactions through the delayed inbox in two primary ways:
+
+1. [General child chain messaging](#general-child-chain-messaging)
+2. [Retryable tickets](#retryable-tickets)
+
```
anegg0, line 104:
```suggestion
Arbitrum provides a _delayed inbox_ contract on the parent chain that can deliver arbitrary messages to the child chain. This functionality is important for two reasons:

1. **General cross-chain messaging**: It allows parent chain EOAs or parent chain contracts to send messages or transactions to the child chain. This is critical for bridging assets (other than the chain's native token) and performing cross-chain operations.
2. **Censorship resistance**: It ensures the Arbitrum chain remains censorship-resistant, even if the Sequencer misbehaves or excludes certain transactions. See [Bypassing the Sequencer](/how-arbitrum-works/02-transaction-lifecycle.mdx#bypassing-the-sequencer) for more details.

Users can send child chain transactions through the delayed inbox in two primary ways:

1. [General child chain messaging](#general-child-chain-messaging)
2. [Retryable tickets](#retryable-tickets)

```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+function sendL2MessageFromOrigin(
+    bytes calldata messageData
+) external whenNotPaused onlyAllowed returns (uint256);
+```
+
+### Unsigned messages
+
+Unsigned messages allow a Parent chain sender to specify transaction parameters without an EOA signature. Because there is no signature, **the sender's address must be aliased on the child chain** (see the [Address aliasing](#address-aliasing) section for the rationale.). The delayed inbox provides four main methods for unsigned messages, divided based on whether the sender is an EOA or a contract and whether it includes parent chain funds:
+
+1. **Unsigned from EOAs**: These methods incorporate a none for replay protection, similar to standard EOA-based transaction on Ethereum.
```
anegg0, line 156:
```suggestion
Unsigned messages allow a Parent chain sender to specify transaction parameters without an EOA signature. Because there is no signature, **the sender's address must be aliased on the child chain** (see the [Address aliasing](#address-aliasing) section for the rationale.). The delayed inbox provides four main methods for unsigned messages, divided based on whether the sender is an EOA or a contract and whether it includes parent chain funds:

1. **Unsigned from EOAs**: These methods incorporate a none-for replay protection, similar to standard EOA-based Ethereum transactions.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 :::
 
-On success, the `To` address receives the escrowed callvalue, and any unused gas is returned to ArbOS's gas pools. On failure, the callvalue is returned to the escrow for the future redeem attempt. In either case, the network fee was paid during the scheduling tx, so no fees are charged and no refunds are made.
+## Retryable tickets
+
+Retryable tickets are Arbitrum's canonical method for creating parent to child chain messges, i.e., parent chain transactions that initiate a message to be executed on the child chain. A retryable can be submitted for a fixed cost (dependent only on its calldata size) paid at the parent chain; its _submission_ on the parent chain is separable/asynchronous with its _execution_ on the child chain. Retryables provide atomicity between the cross chain operations; if the parent chain transaction request submission succeeds (i.e,. does not revert) then the execution of the retryable on the child chain has a strong guarantee to ultimately succeed as well.
```
anegg0, line 228:
```suggestion
Retryable tickets are Arbitrum's canonical method for creating parent-to-child chain messages, i.e., parent-chain transactions that initiate a message to be executed on the child chain. A retryable can be submitted for a fixed cost (dependent only on its calldata size) paid at the parent chain; its _submission_ on the parent chain is separable/asynchronous with its _execution_ on the child chain. Retryables provide atomicity between the cross-chain operations; if the parent chain transaction request submission succeeds (i.e., does not revert) then the execution of the retryable on the child chain has a strong guarantee to ultimately succeed as well.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+Here we walk through the different stages of the lifecycle of a retryable ticket; (1) submission, (2) auto-redemption, and (3) manual redemption.
 
-If a ticket with a callvalue is eventually discarded (cancelled or expired), having never successfully run, the escrowed callvalue will be paid out to a `callValueRefundAddress` account that was specified in the initial submission (step 1).
+#### Submission
 
-:::note Important Notes:
+1. Creating a retryable ticket is initiated with a call (direct or internal) to the `createRetryableTicket` function of the `inbox` [contract](https://github.com/OffchainLabs/nitro-contracts/blob/67127e2c2fd0943d9d87a05915d77b1f220906aa/src/bridge/Inbox.sol). A ticket is guaranteed to be created if this call succeeds. Here, we describe parameters that need to be carefully set. Note that, this function forces the sender to provide a _reasonable_ amount of funds (at least enough to submitting, and _attempting_ to execute the ticket), but that doesn't guarantee a successful auto-redemption.
```
anegg0, line 236:
```suggestion
1. Creating a retryable ticket is initiated with a call (direct or internal) to the `createRetryableTicket` function of the `inbox` [contract](https://github.com/OffchainLabs/nitro-contracts/blob/67127e2c2fd0943d9d87a05915d77b1f220906aa/src/bridge/Inbox.sol). A ticket is guaranteed to be created if this call succeeds. Here, we describe parameters that need to be carefully set. Note that this function forces the sender to provide a _reasonable_ amount of funds (at least enough to submit and _attempt to execute the ticket), but that doesn't guarantee a successful auto-redemption.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+3. Ticket creation causes the `ArbRetryableTx` precompile to emit a `TicketCreated` event containing the `TicketID` on the child chain.
+
+**Ticket submission**
+
+- The user who initiates a parent to child message
+- **Initiating a parent-child message**: a call to `inbox.createRetryableTicket` function that puts the message in the child chain inbox that can be re-executed for some fixed amount of time if it reverts.
+- **Check user's deposit**: logic that checks if the user has enough funds to create a ticket. This is done by checking if the `msg.value` provided by the user is greater than or equal to `maxSubmissionCost` + `l2CallValue` + `gasLimit` \* `maxFeePerGas`
+- **Ticket creation fails**: ticket creation fails and no funds are deducted from the user
+- **Ticket is created**: a ticket is created an added to the retryable buffer on the child chain and funds (`l2CallValue` + `submissionCost`) are deducted to cover the `callValue` from the user and placed into escrow (on the child chain) for later use in redeeming the ticket.
```
anegg0, line 259:
```suggestion
2. Sender's deposit must be enough to make the parent chain submission succeed and for the child chain execution to be _attempted_. If provided correctly, a new ticket with a unique `TicketID` is created and added to the retryable buffer. Also, funds (`submissionCost` + `l2CallValue`) are deducted from the sender and placed into the escrow for later use in redeeming the ticket.
3. Ticket creation causes the `ArbRetryableTx` precompile to emit a `TicketCreated` event containing the `TicketID` on the child chain.

**Ticket submission**

- The user who initiates a parent-to-child message
- **Initiating a parent-child message**: a call to `inbox.createRetryableTicket` function that puts the message in the child chain inbox that can be re-executed for some fixed amount of time if it reverts.
- **Check user's deposit**: logic that checks if the user has enough funds to create a ticket. This is done by checking if the `msg.value` provided by the user is greater than or equal to `maxSubmissionCost` + `l2CallValue` + `gasLimit` \* `maxFeePerGas`
- **Ticket creation fails**: ticket creation fails and no funds are deducted from the user
- **Ticket is created**: a ticket is created and added to the retryable buffer on the child chain and funds (`l2CallValue` + `submissionCost`) are deducted to cover the `callValue` from the user and placed into escrow (on the child chain) for later use in redeeming the ticket.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+#### Automatic redemption
+
+1. It is very important to note that the submission of a ticket on the parent chain is separable/asynchronous from its execution on the child chain, i.e., a successful parent chain ticket creation does not guarantee a successful redemption. Once the ticket is successfully created, the two following conditions are checked: (1) if the user's child chain balance is greater than (or equal to) `maxFeePerGas` \* `gasLimit` **and** (2) if the `maxFeePerGas` (provided by the user in the ticket submission process) is greater than (or equal to) the `l2BaseFee`. If these conditions are both met, the ticket's submission is followed by an attempt to execute it on the child chain (i.e., an **auto-redeem** using the supplied gas, as if the `redeem` method of the [`ArbRetryableTx`](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx) precompile had been called). Depending on how much gas teh sender has provided in Step 1, the ticket's redemption can either (1) immediately succeed or (2) fail. We explain both situations here:
```
anegg0, line 263:
```suggestion
1. It is very important to note that submitting a ticket on the parent chain is separable/asynchronous from its execution on the child chain, i.e., a successful parent chain ticket creation does not guarantee a successful redemption. Once the ticket is successfully created, the two following conditions are checked: (1) if the user's child chain balance is greater than (or equal to) `maxFeePerGas` \* `gasLimit` **and** (2) if the `maxFeePerGas` (provided by the user in the ticket submission process) is greater than (or equal to) the `l2BaseFee`. If these conditions are both met, the ticket's submission is followed by an attempt to execute it on the child chain (i.e., an **auto-redeem** using the supplied gas, as if the `redeem` method of the [`ArbRetryableTx`](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx) precompile had been called). Depending on how much gas the sender has provided in Step 1, the ticket's redemption can either (1) immediately succeed or (2) fail. We explain both situations here:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+#### Manual redemption
+
+1. At this point, _anyone_ can attempt to manually redeem the ticket again by calling [`ArbRetryableTx`](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx)'s `redeem` precompile method, which donates the call's gas to the next attempt. Note that the amount of gas is **not** limited by the original `gasLimit` set during the ticket creation. ArbOS will [enqueue the `redeem`](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L245), which is its own special `ArbitrumRetryTx` type, to its list of redeems that ArbOS [guarantees to exhaust](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L135) before moving on to the next non-redeem transaction in the block its forming. In this manner redeems are scheduled to happen as soon as possible, and will always be in the same block as the tx that scheduled it. Note that the redeem attempt's gas comes from the call to redeem, so there's no chance the block's gas limit is reached before execution.
```
anegg0, line 276:
```suggestion
1. At this point, _anyone_ can attempt to manually redeem the ticket again by calling [`ArbRetryableTx`](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx)'s `redeem` precompile method, which donates the call's gas to the next attempt. Note that the amount of gas is **not** limited by the original `gasLimit` set during the ticket creation. ArbOS will [enqueue the `redeem`](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L245), which is its own special `ArbitrumRetryTx` type, to its list of redeems that ArbOS [guarantees to exhaust](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L135) before moving on to the next non-redeem transaction in the block it's forming. In this manner redeems are scheduled to happen as soon as possible, and will always be in the same block as the tx that scheduled it. Note that the redemption attempt's gas comes from the call to redeem, so there's no chance the block's gas limit is reached before execution.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+2. If the fixed period (one week) elapses withou a successful redeem, the ticket **expires** and will be [automatically discarded](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/retryables/retryable.go#L262), unless some party has paid a fee to [keep the ticket](https://github.com/OffchainLabs/nitro-contracts/blob/a68783436b5105a64f54efe5fbd55174704a7618/src/precompiles/ArbRetryableTx.sol#L41) alive for another full period. A ticket can live indefinitely as long as it is renewed each time before it expires.
```
anegg0, line 277:
```suggestion
2. If the fixed period (one week) elapses without a successful redemption, the ticket **expires** and will be [automatically discarded](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/retryables/retryable.go#L262), unless some party has paid a fee to [keep the ticket](https://github.com/OffchainLabs/nitro-contracts/blob/a68783436b5105a64f54efe5fbd55174704a7618/src/precompiles/ArbRetryableTx.sol#L41) alive for another full period. A ticket can live indefinitely as long as it is renewed each time before it expires.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+#### Avoid losing funds!
+
+If a ticket expires after seven days without being redeemed or re-scheduled to a future date, any message and value (other than the escrowed `callValue`) it carries and could be lost without possibility of being recovered.
+
+On success, the `To` address receives the escrowed `callValue`, and any unused gas is returned to ArbOS's gas pools. On failure, the `callValue` is returned to the escrow for the future `redeem` attempt. In either case, the network fee was paid during the scheduling tx, so no fees are charged and no refunds are made.
+
+Note that during redemption of a ticket, attempts to cancel the same ticket, or to schedule another `redeem` of the same ticket, will revert. In this manner, retryable ticekts are not self-modifying.
+
+If a ticket with a `callValue` is eventually discarded (canceled or expired), having never succesfully run, the escrowed `callValue` will be paid out to a `callValueRefundAddress` account that was specified in the initial submission (Step 1).
```
anegg0, line 291:
```suggestion
If a ticket expires after seven days without being redeemed or re-scheduled to a future date, any message and value (other than the escrowed `callValue`) it carries and could be lost without the possibility of being recovered.

On success, the `To` address receives the escrowed `callValue`, and any unused gas is returned to ArbOS's gas pools. On failure, the `callValue` is returned to the escrow for the future `redeem` attempt. In either case, the network fee was paid during the scheduling tx, so no fees or refunds are charged.

Note that during redemption of a ticket, attempts to cancel the same ticket, or to schedule another `redeem` of the same ticket, will revert. In this manner, retryable tickets are not self-modifying.

If a ticket with a `callValue` is eventually discarded (canceled or expired), having never successfully run, the escrowed `callValue` will be paid out to a `callValueRefundAddress` account that was specified in the initial submission (Step 1).
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+- One can redeem live tickets using the [Arbitrum Retryables Transaction Panel](https://retryable-dashboard.arbitrum.io/tx)
+- The calldata of a ticket is saved on the child chain until it is redeemed or expired
```
anegg0, line 297:
```suggestion
If a redeem is not done at submission or the submission's initial `redeem` fails, anyone can attempt to redeem the retryable again by calling `ArbRetryableTx`'s `redeem` precompile method, which donates the call's gas to the next attempt. ArbOS will [enqueue the redeem](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L245), which is its own special [ArbitrumRetryTx](/build-decentralized-apps/precompiles/02-reference.mdx#arbretryabletx) type, to its list of redeems that ArbOS [guarantees to exhaust](https://github.com/OffchainLabs/nitro/blob/fa36a0f138b8a7e684194f9840315d80c390f324/arbos/block_processor.go#L135) before moving on to the next non-redeem transaction in the block it's forming. In this manner redeems are scheduled to happen as soon as possible, and will always be in the same block as the transaction that scheduled it. Note that the `redeem` attempt's gas comes from the call to `redeem`, so there's no chance the block's gas limit is reached before execution.

- One can redeem live tickets using the [Arbitrum Retryables Transaction Panel](https://retryable-dashboard.arbitrum.io/tx)
- The calldata of a ticket is saved on the child chain until it is redeemed or expired
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+We can build **customized feature messaging** (for example, token bridging) using the messaging systems described in previous sections. In particular, **retryable tickets** power Arbitrum's **canonical token bridge**, which was developed by Offchain Labs. By leveraging retryable tickets under the hood, this token bridge provides a seamless user experience for transferring assets from the parent to child chain. This section is an overview of how the system works.
```
anegg0, line 314:
```suggestion
We can build **customized feature messaging** (for example, token bridging) using the messaging systems described in previous sections. In particular, **retryable tickets** power Arbitrum's **canonical token bridge**, which was developed by Offchain Labs. By leveraging retryable tickets under the hood, this token bridge provides a seamless user experience for transferring assets from the parent-to-child chain. This section is an overview of how the system works.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-## Transacting via the Delayed Inbox
+We now describe some core goals that motivated the design of our bridging system.
 
-While retryables and Eth deposits _must_ be submitted through the <a data-quicklook-from="delayed-inbox">Delayed Inbox</a>, in principle, _any_ message can be included this way; this is a necessary recourse to ensure the <a data-quicklook-from="arbitrum-chain">Arbitrum chain</a> preserves censorship resistance even if the Sequencer misbehaves (see [The Sequencer and Censorship Resistance](/how-arbitrum-works/03-sequencer.mdx)). However, under ordinary/happy circumstances, the expectation/recommendation is that clients use the delayed inbox only for Retryables and Eth deposits, and transact via the Sequencer for all other messages.
+#### Custom gateway functionality
 
-### Address Aliasing
+For many ERC-20 tokens, "standard" bridging functionality is sufficient, which entails the following: a token contract on a parent chain (i.e., Ethereum) is associated with a "paired" token contract on the child chain (i.e., Arbitrum).
```
anegg0, line 328:
```suggestion
For many `ERC-20` tokens, "standard" bridging functionality is sufficient, which entails the following: a token contract on a parent chain (i.e., Ethereum) is associated with a "paired" token contract on the child chain (i.e., Arbitrum).
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 
-Unsigned messages submitted via the Delayed Inbox get their sender's addressed "aliased": when these messages are executed on L2, the sender's address —i.e., that which is returned by `msg.sender` — will _not_ simply be the L1 address that sent the message; rather it will be the address's "L2 Alias." An address's L2 alias is its value increased by the hex value `0x1111000000000000000000000000000000001111`:
+Depositing a token entails escrowing some amount of the token in a parent chain bridge contract, and minting the same amount at the paired token contract on the child chain. On the child chain, the paired contract behaves much like a normal ERC-20 token contract. Withdrawing entails bruning some amount of the token in the child chain contract, which then can later be claimed from the parent chain bridge contract.
```
anegg0, line 330:
```suggestion
Depositing a token entails escrowing some amount of the token in a parent chain bridge contract, and minting the same amount at the paired token contract on the child chain. On the child chain, the paired contract behaves much like a normal `ERC-20` token contract. Withdrawing entails bridging some of the tokens in the child chain contract, which can then later be claimed from the parent chain bridge contract.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
-```sol
-L2_Alias = L1_Contract_Address + 0x1111000000000000000000000000000000001111
-```
+Many tokens, however, require custom gateway systems, the possibilities of which are hard to generalize, e.g.,:
+
+- Tokens which accrue interest to their holders need to ensure that the interest is dispersed properly across layers, and doesn't simply accrue to the bridge contracts
+- Our cross-domain WETH implementations require tokens be wrapped and unwrapped as they move across layers.
+
+Thus, our bridge architecture must allow not only the standard deposit and withdraw functionalities, but for new, custom gateways to by dynamically added over time.
+
```
anegg0, line 338:
```suggestion
Many tokens, however, require custom gateway systems, the possibilities of which are hard to generalize, e.g.,:

- Tokens which accrue interest to their holders need to ensure that the interest is dispersed properly across layers, and doesn't simply accrue to the bridge contracts
- Our cross-domain `WETH` implementations require tokens to be wrapped and unwrapped as they move across layers.

Thus, our bridge architecture must allow not only the standard deposit and withdraw functionalities but also new, custom gateways to be dynamically added over time.

```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+1. **Asset contracts**: These are the token contracts themselves, i.e., an ERC-20 on the parent chain and it's counterpart on the child chain.
+2. **Gatways**: Pairs of contracts (one on the parent chain, and one on the child chain) that implement a particular type of cross-chain asset bridging.
+3. Routers: Exactly two contracts (one on the parent chain, and one on the child chain) that route each asset to its designated gateway.
+
+![500px-img](../assets/canonical-bridge.svg)
+
+All parent to child token transfers are initiated via the router contract on the parent chain, the `L1GatewayRouter` contract. `L1GatewayRouter` forwards the token's deposit call to the appropriate gateway contract on the parent chain, the `L1ArbitrumGateway` contract. `L1GatewayRouter` is responsible for mapping parent chain token addresses to L1Gateway contracts, thus acting as a parent/child address oracle and ensuring each token corresponds to only one gateway. The `L1ArbitrumGateway` then communicates to its counterpart gateway contract on the child chain, the `L2ArbitrumGateway` contract (typically/expectedly via [Retryable tickets](#retryable-tickets)).
+
```
anegg0, line 355:
```suggestion
All parent-to-child token transfers are initiated via the router contract on the parent chain, the `L1GatewayRouter` contract. `L1GatewayRouter` forwards the token's deposit call to the appropriate gateway contract on the parent chain, the `L1ArbitrumGateway` contract. `L1GatewayRouter` is responsible for mapping parent chain token addresses to L1Gateway contracts, thus acting as a parent/child address oracle and ensuring each token corresponds to only one gateway. The `L1ArbitrumGateway` then communicates to its counterpart gateway contract on the child chain, the `L2ArbitrumGateway` contract (typically/expectedly via [Retryable tickets](#retryable-tickets)).

```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+   Please keep in mind that some older custom gateways might not have `outboundTransferCustomRefund` implemented and `L1GatewayRouter.outboundTransferCustomRefund` does not fallback to `outboundTransfer`. In those cases, please use function `L1GatewayRouter.outboundTransfer`.
+   :::
+
+2. `L1GatewayRouter` looks up `SomeERC20Token`'s gateway, and finds that it's the standard ERC-20 gateway (the `L1ERC20Gateway` contract).
+3. `L1GatewayRouter` calls `L1ERC20Gateway.outboundTransferCustomRefund`, forwarding the appropriate parameters.
+4. `L1ERC20Gateway` escrows the tokens sent and creates a retryable ticket to trigger `L2ERC20Gateway`'s `finalizeInboundTransfer` method on the child chain.
+5. `L2ERC20Gateway.finalizeInboundTransfer` mints the appropriate amount of tokens at the `arbSomeERC20Token` contract on the child chain.
+
+:::note
+`arbSomeERC20Token` is an instance of [StandardArbERC20](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/arbitrum/StandardArbERC20.sol), which includes `bridgeMint` and `bridgeBurn` methods only callable by the L2ERC20Gateway.
```
anegg0, line 382:
```suggestion
1. A user calls `L1GatewayRouter.outboundTransferCustomRefund` (with `SomeERC20Token`'s parent chain address as an argument).
   :::warning
   Please keep in mind that some older custom gateways might not have `outboundTransferCustomRefund` implemented and `L1GatewayRouter.outboundTransferCustomRefund` does not fall back to `outboundTransfer`. In those cases, please use function `L1GatewayRouter.outboundTransfer`.
   :::

2. `L1GatewayRouter` looks up `SomeERC20Token`'s gateway, and finds that it's the standard ERC-20 gateway (the `L1ERC20Gateway` contract).
3. `L1GatewayRouter` calls `L1ERC20Gateway.outboundTransferCustomRefund`, forwarding the appropriate parameters.
4. `L1ERC20Gateway` escrows the tokens sent and creates a retryable ticket to trigger `L2ERC20Gateway`'s `finalizeInboundTransfer` method on the child chain.
5. `L2ERC20Gateway.finalizeInboundTransfer` mints the appropriate amount of tokens at the `arbSomeERC20Token` contract on the child chain.

:::note
`arbSomeERC20Token` is an instance of [StandardArbERC20](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/arbitrum/StandardArbERC20.sol), which includes `bridgeMint` and `bridgeBurn` methods only callable by the L2ERC20Gateway.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
 :::
 
-The Arbitrum protocol's usage of L2 Aliases for L1-to-L2 messages prevents cross-chain exploits that would otherwise be possible if we simply reused the same L1 addresses as the L2 sender; i.e., tricking an L2 contract that expects a call from a given contract address by sending retryable ticket from the expected contract address on L1.
+### The Arbitrum generic-custom gateway
 
-If for some reason you need to compute the L1 address from an L2 alias on chain, you can use our `AddressAliasHelper` library:
+Just because a token has requirements beyond what are offered via the standard ERC-20 gateway, that doesn't necessarily mean that a unique gateway needs to be tailor-made for the token in question. Our generic-custom gateway is designed to be flexible enough to be suitable for most (but not necessarily all) custom fungible token needs. As a general rule:
```
anegg0, line 387:
```suggestion
Just because a token has requirements beyond what are offered via the standard `ERC-20` gateway, that doesn't necessarily mean that a unique gateway needs to be tailor-made for the token in question. Our generic-custom gateway is designed to be flexible enough to be suitable for most (but not necessarily all) custom fungible token needs. As a general rule:
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
-    _;
-}
-```
+**If your custom token has the ability to increase its supply (i.e., mint) directly on the child chain, and you want the child chain-minted tokens be withdrawable back to the parent chain and recognized by the parent chain contract, it will probably require its own special gateway. Otherwise, the generic-custom gateway is likely the right solution for you!**
+
+Some examples of token features suitable for the generic-custom gateway:
+
+- A child chain token contract upgradable via a proxy
+- A child chain token contract that includes address whitelisting/blacklisting
+- The deployer determines the address of the child chain token contract
```
anegg0, line 395:
```suggestion
**If your custom token has the ability to increase its supply (i.e., mint) directly on the child chain, and you want the child chain-minted tokens to be withdrawable back to the parent chain and recognized by the parent chain contract, it will probably require its own special gateway. Otherwise, the generic-custom gateway is likely the right solution for you!**

Some examples of token features suitable for the generic-custom gateway:

- A child chain token contract upgradable via a proxy
- A child chain token contract that includes address whitelisting/blacklisting
- The deployer determines the address of the child chain token contract
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+#### Setting up your token with a generic custom gateway
+
+Follow the following steps to get your token set up to use the generic-custom gateway. You can also find more detailed instructions in the page [How to bridge tokens via Arbitrum's genric-custom gateway](/build-decentralized-apps/token-bridging/bridge-tokens-programmatically/03-how-to-bridge-tokens-generic-custom.mdx).
+
+**Pre-requisites**
+Your token on a parent chain should conform to teh [ICustomToken](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/ICustomToken.sol) interface (see [TestCustomTokenL1](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/test/TestCustomTokenL1.sol) for an example implementation). Crucially, it must have an `isArbitrumEnabled` method in its interface.
```
anegg0, line 402:
```suggestion
Follow the following steps to set up your token for the generic-custom gateway. You can also find more detailed instructions on the page [How to bridge tokens via Arbitrum's genric-custom gateway](/build-decentralized-apps/token-bridging/bridge-tokens-programmatically/03-how-to-bridge-tokens-generic-custom.mdx).

**Pre-requisites**
Your token on a parent chain should conform to the [ICustomToken](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/ethereum/ICustomToken.sol) interface (see [TestCustomTokenL1](https://github.com/OffchainLabs/token-bridge-contracts/blob/main/contracts/tokenbridge/test/TestCustomTokenL1.sol) for an example implementation). Crucially, it must have an `isArbitrumEnabled` method in its interface.
```

arbitrum-docs/how-arbitrum-works/10-l1-to-l2-messaging.mdx
```diff
@@ -1,147 +1,437 @@
...
+
+Take our wrapped Ether implementation for example: here, a single WETH contract on the parent chain is connected to a single WETH contract on the child chain. When transferring WETH from one domain to another, the parent/child gateway architecture is used to unwrap the WETH on domain A, transfer the now unwrapped Ether, and then re-wrap it on domain B. This enures that WETH can behave on the child chain the way users are used to it behaving on the parent chain, while ensuring that all WETH tokens are always fully collateralized on the layer in which they reside.
+
+No matter the complexity of a particular token's bridging needs, a gateway can in principle be created to accomodate it within our canonical bridging system.
+
+You can find an example of an implementation of a custom gateway in the page [How to bridge tokens via a custom gateway](/build-decentralized-apps/token-bridging/bridge-tokens-programmatically/04-how-to-bridge-tokens-custom-gateway.mdx).
```
anegg0, line 429:
```suggestion
Note that in the system described above, one pair of gateway contracts handles the bridging of many ERC-20s; i.e., many `ERC-20`s on the parent chain are each paired with their own `ERC-20`s on the child chain via a single gateway contract pairing. Other gateways may well bear different relations with the contracts that they bridge.

Take our wrapped Ether implementation for example: here, a single WETH contract on the parent chain is connected to a single WETH contract on the child chain. When transferring WETH from one domain to another, the parent/child gateway architecture is used to unwrap the `WETH` on domain A, transfer the now unwrapped Ether, and then re-wrap it on domain B. This enures that `WETH` can behave on the child chain the way users are used to it behaving on the parent chain, while ensuring that all `WETH` tokens are always fully collateralized on the layer in which they reside.

No matter the complexity of a particular token's bridging needs, a gateway can in principle be created to accommodate it within our canonical bridging system.

You can find an example of an implementation of a custom gateway on the page [How to bridge tokens via a custom gateway](/build-decentralized-apps/token-bridging/bridge-tokens-programmatically/04-how-to-bridge-tokens-custom-gateway.mdx).
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -15,37 +15,37 @@ _The following document explains the economics and denial-of-service protection
...
+However, Arbitrum technology has not yet achieved its [full promise of being fully decentralized](https://docs.arbitrum.foundation/state-of-progressive-decentralization). Currently, L2 to L1 messaging from Arbitrum One back to Ethereum are verified by a permissioned list of validators. These validators can still challenge invalid withdrawals, but the system prevents anyone outside this list from holding them accountable. This permissioned list of validators limits Arbitrum One and Arbitrum Nova to being categorized as a `Stage 1 Rollup` on the [L2Beat website](https://l2beat.com/scaling/summary), meaning it still has training wheels preventing it from reaching its full potential.
```
anegg0, line 18:
```suggestion
However, Arbitrum technology has not yet achieved its [full promise of being fully decentralized](https://docs.arbitrum.foundation/state-of-progressive-decentralization). Currently, L2 to L1 messaging from Arbitrum One back to Ethereum is verified by a permissioned list of validators. These validators can still challenge invalid withdrawals, but the system prevents anyone outside this list from holding them accountable. This permissioned list of validators limits Arbitrum One and Arbitrum Nova to being categorized as a `Stage 1 Rollup` on the [L2Beat website](https://l2beat.com/scaling/summary), meaning it still has training wheels preventing it from reaching its full potential.
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -15,37 +15,37 @@ _The following document explains the economics and denial-of-service protection
...
+We frequently state that "Arbitrum chains settle their states to a parent chain", and we'll elaborate on what that means. All Arbitrum One transactions can be recreated by reading data from Ethereum L1, as compressed batches of all L2 transactions are frequently posted to Ethereum. Once a batched transaction is included in a finalized block on Ethereum, its history will never be reverted on Arbitrum One. However, when Ethereum receives a batch of transactions, it does not know what the correct result of executing those transactions is. To verify the correct result, there is a separate process that confirms batch correctness on Ethereum: it is called the "assertion."
```
anegg0, line 30:
```suggestion
We frequently state that "Arbitrum chains settle their states to a parent chain", and we'll elaborate on what that means. All Arbitrum One transactions can be recreated by reading data from Ethereum L1, as compressed batches of all L2 transactions are frequently posted to Ethereum. Once a batched transaction is included in a finalized block on Ethereum, its history will never be reverted on Arbitrum One. However, when Ethereum receives a batch of transactions, it does not know what the correct result of executing those transactions is. To verify the correct result, there is a separate process that confirms batch correctness on Ethereum: it is called the "assertion."
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -61,20 +61,23 @@ Another type of validator, called a "proposer," performs additional tasks on top
...
 
 Given that participation in BoLD is permissionless, we recommend that the size of bonds required to participate be high enough to disincentivize malicious actors from attacking Arbitrum One and to mitigate against spam (that would otherwise delay confirmations up to 1 challenge period). High bonding values do not harm decentralization because (1) trustless bonding (or staking) pools can be deployed permissionlessly to open challenges and post assertions, and (2) any number of honest parties of unknown identities can emerge to bond their funds to the correct assertion and participate in the defense of Arbitrum at any time within a challenge. As with the current dispute resolution protocol, there are no protocol level incentives for parties who opt in to participate in validating Arbitrum One with BoLD.
 
-While both of these bonds can be any ERC20 token and be set to any size, we recommend the use of the `WETH` ERC20 token & the following bond sizes for Arbitrum One:
+While both of these bonds can be any ERC-20 token and be set to any size, we recommend the use of the `WETH` ERC-20 token & the following bond sizes for Arbitrum One:
```
anegg0, line 76:
```suggestion
While both of these bonds can be any `ERC-20` token and be set to any size, we recommend the use of the `WETH` `ERC-20` token & the following bond sizes for Arbitrum One:
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -144,13 +155,13 @@ The resource ratio will drive the price of disputes claims, impacting both hones
...
+Although we have made claims with hard numbers about how to price disputes and withdrawal delays in Arbitrum BoLD, we also took a step back and considered the theoretical assumptions we were making. Arbitrum One is a complex protocol used by many groups of people with different incentives. The research team at <a data-quicklook-from="offchain-labs">Offchain Labs</a> has spent considerable effort studying the game theory of validators in optimistic rollup. Honest parties represent everyone with funds onchain, and they have a significant amount to gain by winning the challenge - as they can prevent the loss of their assets rather than losing them.
```
anegg0, line 164:
```suggestion
Although we have made claims with hard numbers about how to price disputes and withdrawal delays in Arbitrum BoLD, we also took a step back and considered the theoretical assumptions we were making. Arbitrum One is a complex protocol used by many groups of people with different incentives. The research team at <a data-quicklook-from="offchain-labs">Offchain Labs</a> has spent considerable effort studying the game theory of validators in optimistic Rollup. Honest parties represent everyone with funds onchain, and they have a significant amount to gain by winning the challenge - as they can prevent the loss of their assets rather than losing them.
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -168,16 +183,16 @@ For Arbitrum One specifically, all eligible entities who wish to be paid this se
...
+For Arbitrum One specifically, 1% (called the “defender’s bounty”) of the confiscated funds from a malicious actor is to be rewarded to honest parties who deposit a challenge bond and post assertions as part of a subchallenge, proportional to the amount that a defender has put up to defend a correct state assertion during the challenge. This bounty applies to all challenges (block challenges, sub challenges and one-step challenges). Note that any gas costs spent by honest parties to defend Arbitrum One during a challenge are 100% refundable by the Arbitrum Foundation. In this model, honest defenders and proposers of Arbitrum One are incentivized to participate while malicious actors stand to lose everything they spent attacking Arbitrum One. We believe chain owners who are interested in adopting BoLD for their own chain should follow a similar approach, described above for Arbitrum One, to incentivize challenge participation (but not necessarily assertion proposing).
```
anegg0, line 186:
```suggestion
For Arbitrum One specifically, 1% (called the “defender’s bounty”) of the confiscated funds from a malicious actor is to be rewarded to honest parties who deposit a challenge bond and post assertions as part of a sub-challenge, proportional to the amount that a defender has put up to defend a correct state assertion during the challenge. This bounty applies to all challenges (block challenges, sub-challenges and one-step challenges). Note that any gas costs spent by honest parties to defend Arbitrum One during a challenge are 100% refundable by the Arbitrum Foundation. In this model, honest defenders and proposers of Arbitrum One are incentivized to participate while malicious actors stand to lose everything they spent attacking Arbitrum One. We believe chain owners who are interested in adopting BoLD for their own chain should follow a similar approach, described above for Arbitrum One, to incentivize challenge participation (but not necessarily assertion proposing).
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -168,16 +183,16 @@ For Arbitrum One specifically, all eligible entities who wish to be paid this se
...
+In this design, defenders are only eligible for the defender's bounty if they deposit a challenge bond (for Arbitrum One, this is either 555 or 79 `ETH`, depending on the level), posted to an on-chain assertion as part of a subchallenge (i.e., not the top-level assertion), and have had their on-chain subchallenge assertion get confirmed by the protocol. For Arbitrum One, the calculation for the defender's bounty is conducted off-chain by the Arbitrum Foundation, and payment will be made via an ArbitrumDAO governance vote (since confiscated funds go to an ArbitrumDAO-controlled address). Honest parties are not automatically rewarded with all the funds seized from malicious actors to avoid creating a situation where honest parties wastefully compete to be the first to make each honest move in the interactive fraud-proof game. Additionally, BoLD resolves disputes by determining which top-level assertion is correct, without necessarily being able to classify every move as “honest” or “malicious” as part of the interactive fraud-proof game without off-chain knowledge.
```
anegg0, line 188:
```suggestion
In this design, defenders are only eligible for the defender's bounty if they deposit a challenge bond (for Arbitrum One, this is either 555 or 79 `ETH`, depending on the level), posted to an on-chain assertion as part of a sub challenge (i.e., not the top-level assertion), and have had their on-chain subchallenge assertion get confirmed by the protocol. For Arbitrum One, the calculation for the defender's bounty is conducted off-chain by the Arbitrum Foundation, and payment will be made via an ArbitrumDAO governance vote (since confiscated funds go to an ArbitrumDAO-controlled address). Honest parties are not automatically rewarded with all the funds seized from malicious actors to avoid creating a situation where honest parties wastefully compete to be the first to make each honest move in the interactive fraud-proof game. Additionally, BoLD resolves disputes by determining which top-level assertion is correct, without necessarily being able to classify every move as “honest” or “malicious” as part of the interactive fraud-proof game without off-chain knowledge.
```

arbitrum-docs/how-arbitrum-works/bold/bold-economics-of-disputes.mdx
```diff
@@ -168,16 +183,16 @@ For Arbitrum One specifically, all eligible entities who wish to be paid this se
...
+This page summarizes the rationale behind choosing bond sizes and the cost of spam prevention in optimistic rollup dispute protocols. We recommend that bond sizes be high enough to discourage challenges from being opened, as malicious parties will always stand to lose when playing the game. As Arbitrum BoLD does not tie disputes to specific addresses, honest parties can have trustless cooperation to resolve disputes if desired. We posit that making the cost of the malicious parties 10x that of the honest party leads to nice economic properties that help us reason about how to price bonds. We describe how a 6.46x ratio (which BoLD as deployed will achieve) represents a pragmatic point in the design space that strikes a balance between concerns about staking costs against concerns about spam. Finally, we look at a high-level game theory discussion of optimistic rollups and argue that solving the verifier's dilemma by incentives to honest validators is an important addition to work towards.
```
anegg0, line 196:
```suggestion
This page summarizes the rationale behind choosing bond sizes and the cost of spam prevention in optimistic rollup dispute protocols. We recommend that bond sizes be high enough to discourage challenges from being opened, as malicious parties will always stand to lose when playing the game. As Arbitrum BoLD does not tie disputes to specific addresses, honest parties can have trustless cooperation to resolve disputes if desired. We posit that making the cost of the malicious parties 10x that of the honest party leads to nice economic properties that help us reason about how to price bonds. We describe how a 6.46x ratio (which BoLD as deployed will achieve) represents a pragmatic point in the design space that strikes a balance between concerns about staking costs and concerns about spam. Finally, we look at a high-level game theory discussion of optimistic rollups and argue that solving the verifier's dilemma by incentives to honest validators is an important addition to work towards.
```

arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.mdx
```diff
@@ -8,32 +8,32 @@ author: leeederek
 sme: leeederek
 ---
 
-This introduction is for those who want to learn about BoLD: a new dispute protocol for Optimistic Rollups that can enable **permissionless validation for Arbitrum chains**. BoLD stands for Bounded Liquidity Delay and is currently deployed on a public testnet for anyone to join and test how challenges will work.
+This introduction is for those who want to learn about BoLD: a new dispute protocol for optimistic rollups that can enable **permissionless validation for Arbitrum chains**. BoLD stands for Bounded Liquidity Delay and is currently deployed on a public testnet for anyone to join and test how challenges will work.
```
anegg0, line 11:
```suggestion
This introduction is for those who want to learn about BoLD: a new dispute protocol for optimistic Rollups that can enable **permissionless validation for Arbitrum chains**. BoLD stands for Bounded Liquidity Delay and is currently deployed on a public testnet for anyone to join and test how challenges will work.
```

arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.mdx
```diff
@@ -8,32 +8,32 @@ author: leeederek
...
+- BoLD, an acronym for Bounded Liquidity Delay, is a new challenge resolution protocol for Arbitrum chains that enables permissionless validation by mitigating the risks of delay attacks against [Optimistic rollups like Arbitrum](/how-arbitrum-works/06-optimistic-rollup.mdx). This is possible because BoLD's design ensures disputes will be resolved within a fixed time window, currently set to equal one challenge period (~6.4 days) for Arbitrum One and Arbitrum Nova. If there is a dispute, BoLD guarantees the maximum total time to be equal to two challenge periods (one for raising disputes, one for resolving disputes), a two day grace period for the Security Council to intervene if necessary, and a small delta for computing challenges.
```
anegg0, line 20:
```suggestion
- BoLD, an acronym for Bounded Liquidity Delay, is a new challenge resolution protocol for Arbitrum chains that enables permissionless validation by mitigating the risks of delay attacks against [Optimistic rollups like Arbitrum](/how-arbitrum-works/06-optimistic-rollup.mdx). This is possible because BoLD's design ensures disputes will be resolved within a fixed time window, currently set to equal one challenge period (~6.4 days) for Arbitrum One and Arbitrum Nova. If there is a dispute, BoLD guarantees the maximum total time to be equal to two challenge periods (one for raising disputes, one for resolving disputes), a two-day grace period for the Security Council to intervene if necessary, and a small delta for computing challenges.
```

arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.mdx
```diff
@@ -180,25 +149,26 @@ Therefore, the BoLD AIP proposes that the honest parties be rewarded 1% of confi
 
 #### Q: Why is $ARB not the bonding token used in BoLD? on Arbitrum One?
 
-Although BoLD supports using an ERC20 token, Ethereum, specifically WETH, was chosen over $ARB for a few reasons:
+Although BoLD supports using an ERC-20 token, Ethereum, specifically WETH, was chosen over $ARB for a few reasons:
```
anegg0, line 152:
```suggestion
Although BoLD supports using an `ERC-20` token, Ethereum, specifically `WETH`, was chosen over `$ARB` for a few reasons:
```

arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.mdx
```diff
@@ -180,25 +149,26 @@ Therefore, the BoLD AIP proposes that the honest parties be rewarded 1% of confi
...
+The topic of economic and incentive models for BoLD on Arbitrum One is valuable and we believe it deserves the full focus and attention of the community via a separate proposal/discussion - decoupled from this proposal to bring BoLD to mainnet. Details around proposed economic or incentive models for BoLD will need continued research and development work, but the deployment of BoLD as-is represents a substantial improvement to the security of Arbitrum even without economic-related concerns being resolved. The DAO may choose, via governance, to fund other parties or change this reimbursement model at any time.
```
anegg0, line 171:
```suggestion
The topic of economic and incentive models for BoLD on Arbitrum One is valuable and we believe it deserves the full focus and attention of the community via a separate proposal/discussion - decoupled from this proposal to bring BoLD to mainnet. Details around proposed economic or incentive models for BoLD will need continued research and development work. Still, the deployment of BoLD as-is represents a substantial improvement to the security of Arbitrum even without economic-related concerns being resolved. The DAO may choose, via governance, to fund other parties or change this reimbursement model at any time.
```

arbitrum-docs/how-arbitrum-works/bold/gentle-introduction.mdx
```diff
@@ -212,13 +182,13 @@ The fee is applied to the total amount bonded over the duration of time that a p
 
 Reimbursement will not be made for off-chain compute costs as we view these to be costs borne by all honest operators, alongside the maintenance and infra costs that regularly arise from running a node.
 
-Our testing has demonstrated that the cost of running a sub-challenge in BoLD, the most computationally-heavy step, on an AWS r5.4xlarge EC2 instance, costs around $2.50 USD (~$1/hour for 1 challenge with 2.5hour duration) using [on-demand prices for US East (N. Virginia)](https://instances.vantage.sh/aws/ec2/r5.4xlarge). Therefore, the additional costs from off-chain compute is assumed to be negligible relative to the regular infra costs of operating a node.
+Our testing has demonstrated that the cost of running a sub-challenge in BoLD, the most computationally-heavy step, on an AWS r5.4xlarge EC2 instance, costs around $2.50 USD (~$1/hour for one challenge with 2.5hour duration) using [on-demand prices for US East (N. Virginia)](https://instances.vantage.sh/aws/ec2/r5.4xlarge). Therefore, the additional costs from off-chain compute is assumed to be negligible relative to the regular infra costs of operating a node.
```
anegg0, line 185:
```suggestion
Our testing has demonstrated that the cost of running a sub-challenge in BoLD, the most computationally-heavy step, on an AWS r5.4xlarge EC2 instance, costs around $2.50 USD (~$1/hour for one challenge with 2.5hour duration) using [on-demand prices for US East (N. Virginia)](https://instances.vantage.sh/aws/ec2/r5.4xlarge). Therefore, the additional costs from off-chain compute are assumed to be negligible relative to the regular infra costs of operating a node.
```

arbitrum-docs/bold/concepts/bold-technical-deep-dive.mdx
```diff
@@ -55,43 +67,48 @@ A helpful mental model for understanding the system is that it uses Ethereum its
...
+Assertions form a chain where there can be forks. For instance, a validator might disagree on the history commitment of block state hashes, which an assertion contains. All <a data-quicklook-from="arbitrum-nitro">Arbitrum Nitro</a> nodes are configured to warn users if they observe an assertion they disagree with posted on-chain. However, suppose a node is configured as a validator and has deposited a bond to the Rollup contract. In that case, that validator can post the correct rival assertion to any invalid one it just observed. The validator will also be able to initiate a challenge by posting a challenge bond and other data to the `ChallengeManager`, signaling it is disputing an assertion.
```
anegg0, line 111:
```suggestion
Assertions form a chain where there can be forks. For instance, a validator might disagree with the history commitment of block state hashes, which an assertion contains. All <a data-quicklook-from="arbitrum-nitro">Arbitrum Nitro</a> nodes are configured to warn users if they observe an assertion they disagree with posted on-chain. However, suppose a node is configured as a validator and has deposited a bond to the Rollup contract. In that case, that validator can post the correct rival assertion to any invalid one it just observed. The validator will also be able to initiate a challenge by posting a challenge bond and other data to the `ChallengeManager`, signaling it is disputing an assertion.
```

arbitrum-docs/bold/concepts/bold-technical-deep-dive.mdx
```diff
@@ -140,7 +159,7 @@ Validators have to fetch all blocks between batch 5 and batch 10 and create a Me
 
 The concept of a history commitment is at the core of challenges and BoLD itself.
 
-The validators above provide a Merkle proof of their commitment to some history. In this case, all the Arbitrum block hashes from batch 5 to batch 10. Using this tree, validators can narrow their disagreement to a single block using Merkle proofs by iteratively bisecting the set of leaves and creating sub-challenge edges which have their own history commitments to each half of the tree.
+The validators above provide a Merkle proof of their commitment to some history. In this case, all the Arbitrum block hashes from batch 5 to batch 10. Using this tree, validators can narrow their disagreement to a single block using Merkle proofs by iteratively bisecting the merkle root and creating edges which have their own history commitments to each half of the tree.
```
anegg0, line 162:
```suggestion
The validators above provide a Merkle proof of their commitment to some history. In this case, all the Arbitrum block hashes from batch 5 to batch 10. Using this tree, validators can narrow their disagreement to a single block using Merkle proofs by iteratively bisecting the Merkle root and creating edges that have their own history commitments to each half of the tree.
```

arbitrum-docs/bold/concepts/bold-technical-deep-dive.mdx
```diff
@@ -194,7 +213,7 @@ Once an edge has a total on-chain timer greater than or equal to a challenge per
 
 ### Bonding in challenges
 
-To create a challenge, there must be a fork in the Arbitrum assertion chain smart contract. A validator that wishes to initiate a challenge must then post an “edge” claiming a history of block hashes from the previous assertion to the claimed assertion they believe is correct. To do so, they need to put up some value called a "challenge bond". Note that to open a new assertion-level challenge, the challenge bond is equal to the assertion bond for <a data-quicklook-from="arbitrum-one">Arbitrum One</a>.
+To create a challenge, there must be a fork in the Arbitrum assertion chain smart contract. A validator that wishes to initiate a challenge must then post an “edge” claiming a history of block hashes from the previous assertion to the claimed assertion they believe is correct. To do so, they need to put up some value called a "challenge bond". Note that to open a new assertion-level challenge, no challenge bond needs to be posted. This is because top-level assertions already contain enough information to uniquely determine their corresponding challenge edge (which contain a hash of this history), and have already been bonded on.
```
anegg0, line 216:
```suggestion
To create a challenge, there must be a fork in the Arbitrum assertion chain smart contract. A validator that wishes to initiate a challenge must then post an “edge” claiming a history of block hashes from the previous assertion to the claimed assertion they believe is correct. To do so, they need to put up some value called a "challenge bond". Note that to open a new assertion-level challenge, no challenge bond needs to be posted. This is because top-level assertions already contain enough information to uniquely determine their corresponding challenge edge (which contains a hash of this history), and have already been bonded on.
```

anegg0 (changes requested):
I left committable comments as editorial nits.

PR #2051, opened 2025-02-07
docs: remove mention of express lane control transferring

leeederek:
because we will not support transferring of timeboost express lane rights in the nitro node software (in the initial release of timeboost), this pr removes all mentions of transferring timeboost rights from our gentle introduction documentation


arbitrum-docs/how-arbitrum-works/timeboost/gentle-introduction.mdx
```diff
@@ -82,12 +82,8 @@ The auction for a round has a closing time that is `auctionClosingSeconds` (def
 
 Before placing a bid in the auction, a party must deposit funds into the Auction Contract. Deposits can be made, or funds can be added to an existing deposit at any time. These deposits are fully withdrawable, with some nominal delay (2 rounds, or 2 minutes by default), so as not to impact the outcome of an existing round. There is no minimum deposit amount, but there is a starting minimum bid of 0.001 ETH (default amount and token) called the "minimum reserve price". The minimum reserve price is set by the chain owner and can be updated at any time up to 30 seconds (default) before the start of the next round to ensure the auction participants will always know the reserve price at least 30 seconds before they must submit their bids. A reserve price can also be set by the chain owner (or by an address designated by the chain owner) as a way to raise the minimum bid as the Auction Contract enforces that the reserve price is never less than the minimum reserve price.
```
anegg0, line 83:
```suggestion
Before placing a bid in the auction, a party must deposit funds into the Auction Contract. Deposits can be made, or funds can be added to an existing deposit at any time. These deposits are fully withdrawable, with some nominal delay (2 rounds or 2 minutes by default), so as not to impact the outcome of an existing round. There is no minimum deposit amount, but there is a starting minimum bid of 0.001 ETH (default amount and token) called the "minimum reserve price". The chain owner sets the minimum reserve price, which can be updated at any time up to 30 seconds (default) before the start of the next round to ensure the auction participants will always know the reserve price at least 30 seconds before they must submit their bids. A reserve price can also be set by the chain owner (or by an address designated by the chain owner) as a way to raise the minimum bid, as the Auction Contract enforces that the reserve price is never less than the minimum reserve price.
```

anegg0 (approved):
I added optional modifications, LGTM

PR #2099, opened 2025-02-21
update --wasm-file command

Jason-W123:
In our cargo-stylus new version, this command is updated from `--wasm-file-path` to `--wasm-file`


joshuacolvin0 (approved):
LGTM

PR #2104, opened 2025-02-26
update chain parameters for bond amount post-BoLD

leeederek:
This small PR updates the chain parameters table under "build decentralized apps" for the new bond requirement for bold.


anegg0 (approved):
lgtm

PR #2120, opened 2025-03-03
add node dependencies install docs link to sqm docs

Jason-W123:
In sqm docs, we don't ask users to install needed packages, which will cause build failed.


anegg0 (approved):
LGTM

PR #2127, opened 2025-03-05
docs: vercel-redirects-pt1

pete-vielhaber:
- Updated vercel.json with Orbit and How Arbitrum Works redirects
- Applied some tidying by moving Orbit, HAW, Stylus, etc., redirects altogether


vercel.json
```diff
@@ -11,22 +11,22 @@
...
+      "destination": "/how-arbitrum-works/interactive-fraud-proofs",
       "permanent": false
     },
     {
-      "source": "/(launch-orbit-chain/orbit-gentle-introduction/?)",
-      "destination": "/launch-orbit-chain/a-gentle-introduction",
-      "permanent": true
+      "source": "/(how-arbitrum-works/fraud-proofs/challenge-manager/?)",
+      "destination": "/how-arbitrum-works/interactive-fraud-proofs",
+      "permanent": false
```
fionnachan, line 21: https://storychief.io/blog/301-302-307-308-redirect

i think we should use `"permanent": true`?

https://vercel.com/docs/edge-network/redirects#best-practices-for-implementing-redirects

pete-vielhaber: Based on feedback from @TucksonDev we are using `false` for consistency. I have no strong opinion one way or the other. It does seem that with the amount of external links using old URL's that are static, it may be wise to make the switch.

TucksonDev: Yeah, we can revisit, but since we are changing docs frequently at the beginning we thought of using "permanent: false" to avoid caching the redirections in the reader's clients.

PR #2129, opened 2025-03-05
Add LaTex Math equations display

anegg0:
This PR adds [Katex](https://docusaurus.io/docs/markdown-features/math-equations) to our Docusaurus instance to display Math equations properly.


pete-vielhaber (approved):
UH-maz-ing - tested with some equations that are pending to be added

PR #2130, opened 2025-03-05
adds support for route change detection

douglance:
when the user commits code:
 1. Checks vercel.json and validates its state
 2. Detects file moves using git diff on staged changes
 3. For each moved markdown file:
    - Converts file paths to URL paths (stripping extensions and prefixes)
    - Checks for existing matching redirects
    - Adds new redirects if needed (always non-permanent)
 4. If redirects are added:
    - Updates vercel.json
    - Requires manual review and staging before continuing


.husky/pre-commit
```diff
@@ -14,4 +14,12 @@ if git diff --cached --name-only | grep -q "src/resources/globalVars.js"; then
   yarn format
 fi
 
+# Check for missing redirects
+yarn ts-node scripts/check-redirects.ts
+CHECK_REDIRECTS_EXIT_CODE=$?
+
+if [ $CHECK_REDIRECTS_EXIT_CODE -ne 0 ]; then
+  exit 1
```
TucksonDev, line 22: Does the precommit shows a message if we exit with an error here? If not, let's log a message here. If we exit without a message, we risk the user thinking that everything was ok and the content committed.

website/vitest.config.ts
TucksonDev, line 1: Should we create an action to run the tests?

website/scripts/check-redirects.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    return true;
+  }
+
+  private getUrlFromPath(filePath: string): string {
+    // Remove the file extension and convert to URL path
+    const urlPath = filePath
+      .replace(/^(pages|arbitrum-docs)\//, '') // Remove leading directory
+      .replace(/\.mdx?$/, ''); // Remove .md or .mdx extension
```
TucksonDev, line 192: We also need to replace any leading numbers with this format '01-', '02-', ... I'm wondering if Docusaurus has already a util for this (saw [this section](https://docusaurus.io/docs/advanced/routing#generating-and-accessing-routes), which leads me to think that the router lib might have a function for that)

website/scripts/check-redirects.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    // Add new redirects
+    let redirectsAdded = false;
+    for (const { oldPath, newPath } of movedFiles) {
+      const oldUrl = this.getUrlFromPath(oldPath);
+      const newUrl = this.getUrlFromPath(newPath);
+
+      // Check if redirect already exists
+      const existingRedirect = vercelJson.redirects.find(
+        (r: any) => r.source === oldUrl && r.destination === newUrl
```
TucksonDev, line 165: oldUrl might be like this `"/(bold/concepts/public-preview-expectations/?)"`, enclosed in `()` and with a `/?` at the end.

website/scripts/check-redirects.ts
```diff
@@ -0,0 +1,203 @@
...
+          });
+        }
+      }
+    }
+
+    console.log('Moved files:', movedFiles);
+    return movedFiles;
+  }
+
+  private convertToUrl(filePath: string): string {
```
TucksonDev, line 104: It looks like this function and the following 2 are not used in the script.

TucksonDev (changes requested):
Looks good! Added a few comments

anegg0 (approved):
lgtm

.github/workflows/test.yml
```diff
@@ -0,0 +1,35 @@
+name: Tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
TucksonDev, line 7: Should this be `master` ? I'm not familiar with the `[]` syntax, but the main branch in this repo is `master`

website/package.json
```diff
@@ -74,11 +75,12 @@
     "markdown-link-extractor": "^3.1.0",
     "prettier": "^2.8.3",
     "styled-components": "^6.1.15",
-    "ts-node": "^10.9.1",
+    "tsx": "^4.19.3",
```
TucksonDev, line 78: Why do we need to change the package in this PR? If it's not related, I'd personally prefer if we change it in a different PR to avoid introducing potential bugs not related to what this PR is intended for. If it's related, nevermind.

.husky/pre-commit
```diff
@@ -14,4 +14,13 @@ if git diff --cached --name-only | grep -q "src/resources/globalVars.js"; then
   yarn format
 fi
 
+# Check for missing redirects
+yarn tsx scripts/check-redirects.ts
```
TucksonDev, line 18: Should we create a script in `package.json` to run this command manually if we want to?

TucksonDev (changes requested):
Thanks @douglance !
Left a few small comments. Nothing blocking except the test which seems to not be running.

website/package.json
TucksonDev: Thanks for the explanation. I ran the different scripts here and they all seem to work correctly. Vercel is also deploying the site correctly so it should be ok.

TucksonDev (approved):
LGTM, thanks very much! 🚀

PR #2141, opened 2025-03-09
fix nitro devnet flags table

leeederek:
this pull request properly formats the table on https://docs.arbitrum.io/run-arbitrum-node/run-local-full-chain-simulation#optional-parameters


pete-vielhaber:
Since this is simply formatting, I'll go forward with merging

PR #2155, opened 2025-03-12
docs: default-docusaurus-pages

pete-vielhaber:
- Removing Docusaurus default demo(s)


anegg0 (approved):
Looking great to me, thanks for the clean up!

PR #2183, opened 2025-03-22
feat: adding guide on fixing contracts submodule not found"

AaronTan11:
1. Audience: Writing for people building in Stylus and facing errors in running devnode
2. Problem: Contracts submodule not found when running devnode
3. Discovery: Running a git command
4. Document type: This is inside quickstart guide

This PR is for adding a quick guide on fixing contracts module not found.

### DIFF
--- a/arbitrum-docs/stylus/quickstart.mdx
+++ b/arbitrum-docs/stylus/quickstart.mdx
@@ -77,6 +77,13 @@ cd nitro-devnode
 ./run-dev-node.sh
 ```
 
+If you see an error like this run the following command:
+```shell title="Error: Contracts submodule not found."
+git submodule update --init --recursive
+```
+
+Then launch your devnode again.
+
 </details>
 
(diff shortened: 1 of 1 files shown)


anegg0 (approved):

AaronTan11:
I'm just curious but what's `sme review` ?

hkalodner:
SME is subject matter expert 😁

I don't think `git submodule update --init --recursive` should be needed since https://github.com/OffchainLabs/nitro-devnode doesn't make use of submodules. Is there a reason why this would be necessary?

joshuacolvin0:
> SME is subject matter expert 😁
> 
> I don't think `git submodule update --init --recursive` should be needed since https://github.com/OffchainLabs/nitro-devnode doesn't make use of submodules. Is there a reason why this would be necessary?

nitro-devnode did briefly use submodules in the past.  Shouldn't be needed anymore.

PR #2226, opened 2025-04-14
Use older Rust version

joshuacolvin0:
Currently `v3.5.x-backports` branch is being used to build the latest releases since master (v3.6.x) is not stable yet.  This means Rust 1.83.0 or older still needs to be used


anegg0 (approved):
lgtm

PR #2253, opened 2025-04-23
docs: temp-katex-fix

pete-vielhaber:
- KaTex integration for equations is causing some odd behavior when $ are used followed by another $. It turns the line in between into an inline equation
- Using a simple escape in front of unintended $ prevents inline equations from displaying


anegg0 (approved):
lgtm

PR #2310, opened 2025-05-09
add expanded ArbOS 40 audit report

0xallan:
contains all the content from the old report (that was previously added to the docs) plus new content from an "audit extension"

also removing the old report from the repo as to not cause confusion


anegg0 (approved):
lgtm

PR #2324, opened 2025-05-14
docs: tx-to-transaction

pete-vielhaber:
- Enforcing terminology guidance, where applicable for tx, should be transaction


anegg0 (approved):
lgtm

PR #2336, opened 2025-05-16
Fix glossary builder

anegg0:
This PR does a few things:
- Renames the CSS class tied to the glossary builder
- Removes the Notion glossary builder
- Regenerates the glossary using the above changes


pete-vielhaber (approved):

website/static/glossary.json
douglance, line 1: if we remove the generation function that uses this, do we still need it? 

looks like it's still used for some tooltips via Tippy in the quicklooks, though that means we have to maintain data in two places, right?

douglance: if we still want to use Tippy tooltips, it might be possible to have a naming convention for a glossary folder that has a bunch of `.mdx` files that we render in the tooltips based on the name of the file. I'd have to look a little closer at it.

pete-vielhaber: We wouldn't need the Notion doc anymore; However you raise a great point - if this impacts Tippy tooltips, we would need to workaround that. I was not aware of an impact between the two.

anegg0: We still use it, and it is still generated by `website/src/scripts/build-glossary.ts`. Do you think it answers your question, @douglance?

douglance: ahhh we had two similar scripts. my mistake. ty sir!

PR #2373, opened 2025-05-30
docs: run-archive-node-update

pete-vielhaber:
- Fixing an inaccuracy; and linking to the proper material (to avoid duplication - single source of truth)


joshuacolvin0 (approved):
LGTM

PR #2392, opened 2025-06-11
update timeboost historical bid data instructions to be region-change resistant

leeederek:
This pr updates the instructions in our documentation so that users can get the historical bid data regardless of the region the s3 bucket is hosted in


anegg0 (approved):
Replaced some basic admonitions with Vanilla Admonition for better readability. LGTM otherwise.

PR #2421, opened 2025-06-20
docs: updating-precompile-links

pete-vielhaber:
- [old precompile](https://github.com/OffchainLabs/nitro-contracts/tree/main/src/precompiles) links are broken 
- should be updated to [new repo](https://github.com/OffchainLabs/nitro-precompile-interfaces)


anegg0 (approved):
I tested the links, and they're not broken anymore

PR #2487, opened 2025-07-23
Update building nitro docs for go 1.24

eljobe:
With the upcoming 3.7.0 release, we are starting to build Nitro with go version 1.24.5 (but, any version in the 1.24 series should be fine.)


joshuacolvin0 (approved):
LGTM

PR #2491, opened 2025-07-25
docs: nit3556-eip2935

pete-vielhaber:
- Adding clarity around the `blockhash()` call


arbitrum-docs/build-decentralized-apps/arbitrum-vs-ethereum/04-solidity-support.mdx
```diff
@@ -15,12 +15,12 @@ This page compiles a list of functions and properties that return a different re
...
+| ------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `blockhash(x)`     | In most cases, it will return a cryptographically insecure, pseudo-random hash for `x` within the range `block.number - 256 <= x < block.number`. (In rare instances, if an L2 block is not linked to an Ethereum (L1) block, you may get a random value.) If `x` is outside of this range, `blockhash(x)` will return `0`. This includes `blockhash(block.number)`, which always returns `0` just like on Ethereum. The returned hashes do not originate from the parent chain. ⚠️ Arbitrum's child chain block hashes should not be relied on as a secure source of randomness. |
```
leeederek:
```suggestion
| `blockhash(x)`     | In most cases, when provided with an L2 block number, this will return the L2 block hash associated with that number, up to 8192 L2 blocks. In rare cases, if an L2 block is not linked to an Ethereum (L1) block, it will return a cryptographically insecure, pseudo-random hash for `x` within the range `block.number - 256 <= x < block.number`. If `x` is outside of this range *and* is is an L2 block that is not associated with an L1 block, then `blockhash(x)` will return `0` (just like it would on Ethereum). The returned hashes do not originate from the parent chain. ⚠️ Arbitrum's child chain block hashes should not be relied on as a secure source of randomness. |
```

tsahee (changes requested):
AFAIK blockhash(x) did not change and neither should it's documentation

PR #2616, opened 2025-08-22
docs: tw536-anytrust-native-token-language

pete-vielhaber:
- AnyTrust for Arbitrum chain:
  - Tokens do not need to be natively deployed on L1; some representation needs to be on L1


hkalodner (approved):
LGTM

PR #2641, opened 2025-08-28
update nodejs version to v23

Jason-W123:
When compiling v3.7.0, logs show we need to use `{ node: '^20.19.0 || ^22.12.0 || >=23' }`


Jason-W123:
Draft as we still use v3.6.8 in main docs

PR #2858, opened 2025-11-11
Improve redirect script

anegg0:
This PR improves redirect handling by making them permanent and avoiding circular redirects.
The main objective was to ensure the persistence of links that had been modified twice without affecting performance.


scripts/check-redirects.ts
```diff
@@ -293,24 +334,227 @@ export class RedirectChecker {
...
+   * Detect and report existing redirect chains in the configuration
+   * Returns an array of chains found (for logging/debugging purposes)
+   * Uses index-based lookup for O(1) performance
+   */
+  private detectExistingChains(indices: RedirectIndices): string[][] {
+    const chains: string[][] = [];
+    const processed = new Set<string>();
+
+    // Convert Map keys to array for iteration
+    const sources = Array.from(indices.bySource.keys());
```
EmreDincoglu, line 472: You should be able to directly iterate over a map in JS instead of converting it to an array. Im not sure how it compiles under the hood (and i've barely used Js) but it should be more efficient than copying it over to an array to iterate over

example I found online
    const myMap = new Map([
        ['a', 1],
        ['b', 2],
        ['c', 3]
    ]);

    for (const [key, value] of myMap) {
        console.log(`Key: ${key}, Value: ${value}`);
    }

EmreDincoglu: im silly this is TS which has the map.forEach which should allow you to skip copying it over to an array

EmreDincoglu (approved):

scripts/check-redirects.ts
anegg0: No pb! Thanks for the review, @EmreDincoglu !

PR #3053, opened 2026-02-02
docs: tw590-fast-block-times

pete-vielhaber:
- Adding configuration for fast block times for Arbitrum chains

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -509,6 +509,11 @@ const sidebars = {
                       id: 'launch-arbitrum-chain/configure-your-chain/common/validation-and-security/bold-adoption-for-arbitrum-chains',
                       label: `Adopt BoLD`,
                     },
+                    {
+                      type: 'doc',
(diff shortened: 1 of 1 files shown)


Jason-W123:
Hey, I think this doc incorrectly says block time has relationship with `sequencerInboxMaxTimeVariation`. That parameter lives on the L1 sequencerInbox contract and controls the allowed time bounds for batch submission — it has nothing to do with L2 block production speed. We already have a [doc](https://docs.arbitrum.io/launch-arbitrum-chain/configure-your-chain/common/validation-and-security/batch-posting-assertion-control#sequencer-max-time-variation) introduces this.
The actual parameter that controls block time is `--execution.sequencer.max-block-speed` (default 250ms). This should be the focus of the doc.
Maybe we need to remove all `sequencerInboxMaxTimeVariation` contents on this doc.

pete-vielhaber:
We already have covered as mention in comment above

PR #3197, opened 2026-04-08
rename notion-export-csv to notion-export-qa-csv for clarity

Jason-W123:
The script specifically exports QA data from Notion for Inkeep, not general Notion content. The new name makes this purpose obvious.


pete-vielhaber (approved):
LGTM

PR #3228, opened 2026-04-22
docs: orbit-actions-rename

pete-vielhaber:
- Updating `orbit-actions` refs
  - chain-actions
- Also renaming L3 pages (launch-arbitrum-chain)
  - Renamed to make them more identifiable to what they actually are
  - Also re-wrote the first paragraph on Launch your first rollup (renamed) to explain why this page is using defaults, etc.


anegg0 (approved):
LGTM

PR #3230, opened 2026-04-23
update and fix signermask

Jason-W123:
signermask field is already removed on [nitro pr 2427](https://github.com/OffchainLabs/nitro/pull/2427)


pete-vielhaber (approved):
LGTM

PR #3256, opened 2026-05-01
docs: styling-currency-symbols

pete-vielhaber:
- After some discussion we have decided to change the current inline code styling for currency symbols with bold instead.
- Post CSS update a little while back it has caused some visual "ugliness"


anegg0 (approved):
LGTM

PR #3284, opened 2026-05-15
Improve LLM/AI agent access to docs

hkalodner:
## Summary

- **Tighten sitemap** so it indexes the same canonical pages as `llms.txt` — partials, auto-generated category pages, and SDK API internals are now excluded from both indexes via a shared pattern list. Drops sitemap entries from 503 to 275.
- **Surface `llms.txt` to AI agents on every page**: a screen-reader-only div in the layout points to `/llms.txt`, `/llms-full.txt`, and the `.md` variant of any URL. A remark plugin prepends a `> For a complete page index, fetch <https://docs.arbitrum.io/llms.txt>` blockquote to every per-page `.md` and every section in `llms-full.txt`.
- **Content negotiation via Vercel**: requests with `Accept: text/markdown` are rewritten to the `.md` file (`/foo` → `/foo.md`, `/` → `/index.md`). Paths containing dots pass through unchanged so static assets and direct `.md` requests aren't affected. `Vary: Accept` is set on the same routes so CDN/browser caches don't conflate the two representations.

## Test plan

- [ ] On a preview deploy, fetch a doc page with `curl -H 'Accept: text/markdown' <url>` and confirm it returns the markdown body
- [ ] Same request without that header returns the HTML page
- [ ] Each preview-deploy doc page contains `For a complete page index` in the markdown output (`curl <url>.md | head -3`)
- [ ] Sitemap at `<url>/sitemap.xml` no longer contains `/partials/`, `/category/`, or `/sdk/(assetBridger|dataEntities|inbox|message|utils)/` entries
- [ ] View page source on a deployed page and confirm the `sr-only` notice div is present and not visible

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -61,6 +61,22 @@ const handleInkeepEvent = (event) => {
   }
 };
 
+// Routes that exist in the Docusaurus build but aren't standalone, indexable pages.
+// Shared between the sitemap and llms.txt so both indexes stay in sync.
+const nonCanonicalRoutePatterns = [
+  '/sdk/assetBridger/**',
+  '/sdk/dataEntities/**',
+  '/sdk/inbox/**',
+  '/sdk/message/**',
+  '/sdk/utils/**',
+  '/hosted-pdfs/**',
+  // Partials are imported into other pages, not standalone content.
+  // Docusaurus generates routes for them anyway.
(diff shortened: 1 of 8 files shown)


anegg0 (approved):
LGTM

PR #3410, opened 2026-06-24
Update Stylus how-to guides for SDK 0.10.7

anegg0:
**Stylus docs update** — part of a small set of PRs (tagged `stylus`) bringing `docs/stylus/` in line with the current stack (stylus-sdk / cargo-stylus **0.10.7**, Nitro **v3.11.0**).

- `how-tos/exporting-abi.mdx` — remove phantom CLI (`export-abi constructor`, `--license`, `--pragma`); document `cargo stylus constructor`.
- `how-tos/using-constructors.mdx` — fix the event-emission API; add `--constructor-value` for payable constructors.
- `how-tos/trait-based-composition.mdx` — add required `#[public]` on interface traits; modernize the selector example.

Examples verified against the live toolchain (compiled/deployed on a local Nitro dev node where relevant) and conceptual claims against the Nitro source. Builds clean (`yarn build`, strict `onBrokenLinks`). Version strings are literals here; they are centralized into `globalVars.js` in a follow-up PR.


docs/stylus/how-tos/exporting-abi.mdx
EmreDincoglu: I cant get this code to work without passing in `&mut self` (I tried adding the suggest change but it just removes a blank line and adds this one)
```suggestion
    pub fn new(&mut self, owner: Address, initial_value: U256) {
```

docs/stylus/how-tos/exporting-abi.mdx
```diff
@@ -281,14 +258,14 @@ impl Token {
 }
```
EmreDincoglu, line 266: same with the code here, im unable to get this working without adding this extra code here
```suggestion
sol! {
error InsufficientBalance(address account, uint256 requested, uint256 available);
error Unauthorized(address caller);
error InvalidAmount();
}
#[derive(SolidityError)]
pub enum TokenError {
    InsufficientBalance(InsufficientBalance),
    Unauthorized(Unauthorized),
    InvalidAmount(InvalidAmount)
}

#[public]
impl Token {
pub fn transfer(&mut self, to: Address, amount: U256) -> Result<(), TokenError> {
let sender = self.vm().msg_sender();
let balance = self.balances.get(sender);
if balance < amount {
return Err(TokenError::InsufficientBalance(InsufficientBalance {
account: sender,
requested: amount,
available: balance,
}));
}
// Transfer logic
Ok(())
}
}
```

tldr 
`#[derive(SolidityError)]
pub enum TokenError {
    InsufficientBalance(InsufficientBalance),
    Unauthorized(Unauthorized),
    InvalidAmount(InvalidAmount)
}`

on top of changing the result to receiving an enum, im not fully sure why it doesnt work but claude says it has to do with it not being able to return into byes without this enum pattern with the `#[derive(SolidityError)]` above it

EmreDincoglu (commented):
I will continue the review on monday, but as of right now either stylus is just super hard to use or there is just a lot of changes that make previous things work no longer work

EmreDincoglu:
the updated code all seems to work, given the things I mentioned in the parts that were not updated (I went through the doc to learn and test thats how I caught it)

anegg0:
> I will continue the review on monday, but as of right now either stylus is just super hard to use or there is just a lot of changes that make previous things work no longer work

Yes, same assessment here. There are quite a few things that have changed lately.

docs/stylus/how-tos/exporting-abi.mdx
```diff
@@ -278,17 +255,44 @@ impl Token {
         // Transfer logic
         Ok(())
     }
```
EmreDincoglu, line 265: This gotta be deleted, I cant add a suggestion due to the GH bug I was talking abt, I tried doing the suggestion removing all the lines in this chunk makes GH think im trying to delete other lines

docs/stylus/how-tos/exporting-abi.mdx
anegg0: The edits I made should work. Could you confirm?

docs/stylus/how-tos/exporting-abi.mdx
anegg0: I made a few more changes; this should compile.
Please let me know if that's acceptable.

EmreDincoglu (approved):
I think this is good to go

PR #3491, opened 2026-07-29
docs: tw-767-token-bridge-troubleshooting

pete-vielhaber:
Thank you for contributing to our docs!

Please fill out the form below to ensure your doc gets quickly approved and merged.

## Description
- Adding a troubleshooting ref/guide for deploying/operating the token bridge for Arbitrum chains

## Document type

- [ ] Gentle introduction
- [ ] Quickstart
- [X] How-to
- [ ] Concept
- [ ] FAQ
- [ ] Troubleshooting
- [X] Reference
- [ ] Third-party content
- [ ] Codebase changes
- [ ] Not applicable

## Checklist
- [X] I have read the [CONTRIBUTE.md](../CONTRIBUTE.md) guidelines
- [X] My changes follow the style conventions outlined in CONTRIBUTE.md
- [X] I have used sentence-case for titles and headers
- [X] I have used descriptive link text (not "here" or "this")
- [X] I have separated procedural from conceptual content where appropriate
- [X] I have tested my changes locally with `yarn start` or `yarn build`
- [X] My code follows the existing code style and conventions
- [X] I have added/updated frontmatter for new documents
- [X] I have checked for broken links
- [X] I have verified that my changes don't break the build
- Third-party docs only: Do you agree to the third-party content policy outlined within [CONTRIBUTE.md](../CONTRIBUTE.md)?
  - [ ] Yes
  - [X] Not applicable


docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+Most routing problems in this guide trace back to a single behavior: the <a data-quicklook-from="gateway-router">gateway router</a> falls back to a default gateway instead of failing when a token has no registration. Start with the following section, because the symptoms in [Bridged USDC routes to the standard gateway](#bridged-usdc-routes-to-the-standard-gateway) and [Custom gateway registration fails](#custom-gateway-registration-fails) are both consequences of it.
+
+## How the canonical router picks a gateway
+
+Each side of the bridge has a router (`L1GatewayRouter` on the <a data-quicklook-from="parent-chain">parent chain</a>, `L2GatewayRouter` on the <a data-quicklook-from="child-chain">child chain</a>) holding an `l1TokenToGateway` mapping from a parent chain token address to the gateway that handles it. When you bridge a token, the router calls `getGateway`, which resolves in this order:
```
TucksonDev: `l1TokenToGateway` only contains entries for tokens that have a custom gateway set for them. The text implies that the reader knows that (I think, based on what I've read so far) so maybe this should be stated here.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+3. If the resolved gateway is explicitly disabled, return the zero address.
+
+<VanillaAdmonition type="warning" title="Unregistered tokens fail silently, not loudly">
+
+Step 2 is the reason misrouting is so hard to spot. A token you believe is registered to a <a data-quicklook-from="custom-gateway">custom gateway</a>—but isn't—still bridges successfully through the standard gateway, minting a _different_ child chain token than the one you deployed. Nothing reverts and no event signals the mistake. If a bridge transfer succeeds but the tokens land at an unexpected address, suspect a missing registration before you suspect a bug.
+
+</VanillaAdmonition>
+
+The router also rewrites the calldata it forwards. It wraps the original sender with `GatewayMessageHandler.encodeFromRouterToGateway(msg.sender, _data)` so the gateway can recover the true sender. Gateways detect this by checking `isRouter(msg.sender)`. That distinction is what makes direct gateway calls behave differently from router calls, as described in [`outboundTransfer` reverts or moves the wrong tokens](#outboundtransfer-reverts-or-moves-the-wrong-tokens).
```
TucksonDev: This is only possible on L2->L1 transfers, iirc. But in any case, I think this paragraph doesnt provide much value to the reader at this point, and might make things more confusing.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+author: pete-vielhaber
+sme: jose-franco
+user_story: As an Arbitrum chain operator, I need to diagnose and resolve common token bridge deployment and registration problems so that tokens bridge through the gateways I intended.
+content_type: troubleshooting
+---
+
+This guide covers problems that operators encounter when deploying, verifying, and registering contracts for the canonical <a data-quicklook-from="arb-token-bridge">token bridge</a>. For the deployment procedure itself, see [Deploy a token bridge using the Chain SDK](/launch-arbitrum-chain/deploy/token-bridge.mdx). To learn how the bridge is designed, see [Token bridging](/how-arbitrum-works/deep-dives/token-bridging.mdx).
+
+Most routing problems in this guide trace back to a single behavior: the <a data-quicklook-from="gateway-router">gateway router</a> falls back to a default gateway instead of failing when a token has no registration. Start with the following section, because the symptoms in [Bridged USDC routes to the standard gateway](#bridged-usdc-routes-to-the-standard-gateway) and [Custom gateway registration fails](#custom-gateway-registration-fails) are both consequences of it.
```
TucksonDev: I think this text "Most routing problems in this guide trace back to a single behavior" is a bit misleading. I was reading the document expecting these kind of issues, but didn't see any until the fourth section. Maybe it can be reworded or the "how the canonical router picks a gateway" section can be moved below, just before the issues that are related to this.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+<VanillaAdmonition type="warning" title="Unregistered tokens fail silently, not loudly">
+
+Step 2 is the reason misrouting is so hard to spot. A token you believe is registered to a <a data-quicklook-from="custom-gateway">custom gateway</a>—but isn't—still bridges successfully through the standard gateway, minting a _different_ child chain token than the one you deployed. Nothing reverts and no event signals the mistake. If a bridge transfer succeeds but the tokens land at an unexpected address, suspect a missing registration before you suspect a bug.
+
+</VanillaAdmonition>
+
+The router also rewrites the calldata it forwards. It wraps the original sender with `GatewayMessageHandler.encodeFromRouterToGateway(msg.sender, _data)` so the gateway can recover the true sender. Gateways detect this by checking `isRouter(msg.sender)`. That distinction is what makes direct gateway calls behave differently from router calls, as described in [`outboundTransfer` reverts or moves the wrong tokens](#outboundtransfer-reverts-or-moves-the-wrong-tokens).
+
+## Shared WETH and Multicall deployments
```
TucksonDev: I see this section as something to include in a guide that describes how TokenBridges are created, but not as troubleshooting 🤔 .
The admonition at the end of it does seem as the type of content for a troubleshooting guide though.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+Deploying the token bridge creator against a non-Ethereum parent chain fails with output like this:
+
+```console
+Deploying token bridge creator...
+Duplicate definition of Transfer (Transfer(address,address,uint256,bytes), Transfer(address,address,uint256))
+Error: network does not support ENS (operation="getResolver", network="unknown", code=UNSUPPORTED_OPERATION, version=providers/5.7.2)
+```
+
+### What causes this error
```
TucksonDev: I'd probably summarize this section a bit more. It looks like it contains implementation details that are not needed for a reader who's troubleshooting the issue. A simply "check BASECHAIN_WETH (or ROLLUP_ADDRESS if GAS_LIMIT_... is unset) because those are the only addresses that is directly passed to ethers v5" would suffice IMO. 
Actually, you probably don't even need to mention ethers v5

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+```
+
+Seeing the ENS error instead of one of those means the variable is **set but malformed**. Two values reach ethers as addresses:
+
+- `BASECHAIN_WETH`, passed straight into `deployL1TokenBridgeCreator` and on to `setTemplates`. Common defects include placeholder text, a wrapping pair of quotes, trailing whitespace, a missing `0x` prefix, and a truncated address.
+- `ROLLUP_ADDRESS`, used by `RollupAdminLogic__factory.connect(rollupAddress, l1Provider)`. This is only read when you leave `GAS_LIMIT_FOR_L2_FACTORY_DEPLOYMENT` unset, because that's the path that estimates the gas limit instead.
+
+### Resolution
+
+<!-- SME review: step 2 below is inferred from the contract source (setTemplates stores l1Weth, and deployTokenBridgeCreator.ts rejects an empty BASECHAIN_WETH), not from an upstream doc that states the requirement. Please confirm the guidance for a parent chain with no canonical wrapped native token—specifically whether deploying a WETH9-compatible contract is what we want to recommend, and whether an ETH-based chain can instead be deployed with a placeholder WETH address. -->
```
TucksonDev: This comment probably needs to be removed

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+1. **Verify `BASECHAIN_WETH` is a bare, checksummed hex address.** No quotes, no whitespace, `0x` prefixed, 42 characters.
+2. **If your parent chain has no wrapped native token, deploy one first.** The creator stores a parent chain **WETH** address and the deploy script rejects an empty `BASECHAIN_WETH`, so a chain without a canonical wrapped native contract needs one deployed before the bridge can be created. A `WETH9`-compatible contract is the usual choice.
```
TucksonDev: I don't think this is needed for custom gas token chains 🤔

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+3. **Set `GAS_LIMIT_FOR_L2_FACTORY_DEPLOYMENT` explicitly.** Approximately `6000000` is a reasonable value. This skips the `ROLLUP_ADDRESS` and `ORBIT_RPC` estimation path, removing a second source of the same error. The script requires either this variable or both `ROLLUP_ADDRESS` and `ORBIT_RPC`.
```
TucksonDev: I wouldn't advice a deployer to do this explicitly. They need to understand what they are doing. So the options they have wrt calculating the gas limits should probably belong in a different kind of guide.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+4. **Confirm the deployer key and RPC belong to the same chain.** A key funded on a different chain produces confusing downstream failures once the ENS problem is cleared.
```
TucksonDev: This is not related to the problem being solved

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+<VanillaAdmonition type="tip" title="The Transfer warning is benign">
```
TucksonDev: Instead of adding this admonition, I would remove the warnning from the output above.

TucksonDev (changes requested):
Reviewed until "network does not support ENS". Will provide feedback internally.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
TucksonDev: With the new changes, I'd suggest changing the text to say something like "we’ll first cover issues related to how to gateway router picks a specific gateway for a token"

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+Most routing problems in this guide trace back to a single behavior: the <a data-quicklook-from="gateway-router">gateway router</a> falls back to a default gateway instead of failing when a token has no registration. Start with the following section, because the symptoms in [Bridged USDC routes to the standard gateway](#bridged-usdc-routes-to-the-standard-gateway) and [Custom gateway registration fails](#custom-gateway-registration-fails) are both consequences of it.
+
+## How the TokenBridge router picks a gateway
+
+Each side of the bridge has a router (`L1GatewayRouter` on the <a data-quicklook-from="parent-chain">parent chain</a>, `L2GatewayRouter` on the <a data-quicklook-from="child-chain">child chain</a>) holding an `l1TokenToGateway` mapping from a parent chain token address to the gateway that handles it.
+
+The mapping isn't a list of every bridgeable token. It holds only the tokens that someone registered to a <a data-quicklook-from="custom-gateway">custom gateway</a>, so an ordinary **ERC-20** token never has an entry. When you bridge a token, the router calls `getGateway`, which resolves in this order:
```
TucksonDev:
```suggestion
The mapping isn't a list of every bridgeable token. It holds only the tokens that someone registered to a <a data-quicklook-from="custom-gateway">custom gateway</a>, so a given **ERC-20** token may or may not have an entry. When you bridge a token, the router calls `getGateway`, which resolves in this order:
```

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+sme: jose-franco
+user_story: As an Arbitrum chain operator, I need to diagnose and resolve common token bridge deployment and registration problems so that tokens bridge through the gateways I intended.
+content_type: troubleshooting
+---
+
+This guide covers problems that operators encounter when deploying, verifying, and registering contracts for the canonical <a data-quicklook-from="arb-token-bridge">token bridge</a>. For the deployment procedure itself, see [Deploy a token bridge using the Chain SDK](/launch-arbitrum-chain/deploy/token-bridge.mdx). To learn how the bridge is designed, see [Token bridging](/how-arbitrum-works/deep-dives/token-bridging.mdx).
+
+Most routing problems in this guide trace back to a single behavior: the <a data-quicklook-from="gateway-router">gateway router</a> falls back to a default gateway instead of failing when a token has no registration. Start with the following section, because the symptoms in [Bridged USDC routes to the standard gateway](#bridged-usdc-routes-to-the-standard-gateway) and [Custom gateway registration fails](#custom-gateway-registration-fails) are both consequences of it.
+
+## How the TokenBridge router picks a gateway
```
TucksonDev: Since this section is different than the others, maybe include something in the title like "Context: How ..." or something like that ("Context" is probably not a good solution). 
Or include something in the other sections like "Problem: ..." or Issue or any similar word.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+2. If it has no entry, fall back to `defaultGateway`—the <a data-quicklook-from="standarderc20-gateway">standard **ERC-20** gateway</a>.
+3. If the resolved gateway is explicitly disabled, return the zero address.
+
+<VanillaAdmonition type="warning" title="Unregistered tokens fail silently, not loudly">
+
+Step 2 is the reason misrouting is so hard to spot. A token you believe is registered to a custom gateway—but isn't—still bridges successfully through the standard gateway, minting a _different_ child chain token than the one you deployed. Nothing reverts and no event signals the mistake. If a bridge transfer succeeds but the tokens land at an unexpected address, suspect a missing registration before you suspect a bug.
+
+</VanillaAdmonition>
+
+## Bridged USDC routes to the standard gateway
```
TucksonDev, line 31: Maybe include a link from the "Bridged USDC" page to this page for troubleshooting.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+You deployed a bridged **USDC** gateway following [the bridged USDC standard](/launch-arbitrum-chain/integrations/bridged-usdc.mdx), but the SDK and bridge UI keep sending **USDC** through the standard **ERC-20** gateway, producing the default bridged token rather than the Circle-standard one.
+
+### What causes this problem
+
+The deployment script's final registration step is conditional. `deployUsdcBridge.ts` performs these operations in order:
```
TucksonDev:
```suggestion
The deployment script's final registration step is conditional. It performs these operations in order:
```

In general, I'd avoid using specific script names here. File names can change from time to time and we probably don't want to have to update multiple docs because of it.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
TucksonDev: Also, since this is troubleshooting an issue related to a specific process that is explained in its own page, consider creating a Troubleshooting section in that guide instead.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+Step 2 is the reason misrouting is so hard to spot. A token you believe is registered to a custom gateway—but isn't—still bridges successfully through the standard gateway, minting a _different_ child chain token than the one you deployed. Nothing reverts and no event signals the mistake. If a bridge transfer succeeds but the tokens land at an unexpected address, suspect a missing registration before you suspect a bug.
+
+</VanillaAdmonition>
+
+## Bridged USDC routes to the standard gateway
+
+You deployed a bridged **USDC** gateway following [the bridged USDC standard](/launch-arbitrum-chain/integrations/bridged-usdc.mdx), but the SDK and bridge UI keep sending **USDC** through the standard **ERC-20** gateway, producing the default bridged token rather than the Circle-standard one.
+
+### What causes this problem
```
TucksonDev, line 35: I think this whole section could be simplified.
Describing every step that the script takes doesn't add much value to the reader. And the description is based on a very specific scenario (the chain owner is an UpgradeExecutor contract), which might not be always the case.
Btw, most operators running a chain are probably aware of how this type of script works (admin actions are either executed or the calldata is printed), so I wouldn't expect many readers to fall into the scenario described here.

I'd simplify this whole section by saying something like "As part of the USDC bridge deployment script, the custom USDC gateway is registered in the TokenBridge router. If that process fails, the TokenBridge won't be able to choose the right gateway for USDC, falling back to the default gateway".

And just continue with the resolution section.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+4. Deploy and initialize the parent and child chain **USDC** gateways.
+5. **Register the gateways with the routers**—only when `ROLLUP_OWNER_KEY` is set.
+6. Configure the child chain gateway as minter with maximum allowance.
+
+Step 5 runs the registration through the `UpgradeExecutor`. Without `ROLLUP_OWNER_KEY`, the script doesn't fail. It writes the registration calldata to `registerUsdcGatewayTx.json` and expects you to execute it separately. If that file is never executed, `l1TokenToGateway` has no entry for **USDC**, and `getGateway` returns `defaultGateway` exactly as described in [How the router picks a gateway](#how-the-tokenbridge-router-picks-a-gateway).
+
+### Resolution
+
+1. **Query the router directly.** Call `getGateway` on `L1GatewayRouter` with your parent chain **USDC** address. If it returns the standard **ERC-20** gateway rather than your **USDC** gateway, registration never landed.
+2. **Execute `registerUsdcGatewayTx.json`.** Submit the stored calldata through your `UpgradeExecutor` using the account holding the executor role.
```
TucksonDev: I'd put this option after the current number (3), so the list would be (1), (3), (2)...

Then, I'd modify the text to say something like: "If your chain owner is the `UpgradeExecutor` contract, make sure an account with the EXECUTOR role sent the transaction to register the gateways, which should've been generated by the USDC bridge deployment script".

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+3. **Check both sides.** Registration crosses chains via a <a data-quicklook-from="retryable-ticket">retryable ticket</a>, so confirm the child chain router resolves correctly too. See [Registration succeeded on one chain only](#registration-succeeded-on-one-chain-only).
+4. **Confirm gateway ownership.** Both gateways must be owned by the `UpgradeExecutor`, with an account or multisig holding the executor role. See [Ownership and access](/launch-arbitrum-chain/operate/ownership-and-access.mdx).
```
TucksonDev: This shouldn't affect this problem. I'd remove this point.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+## Custom gateway registration fails
+
+Registering a token or gateway involves four entry points across two contracts, and they split into two authorization models: token-side calls that the token itself must make, and owner-side calls that only the contract owner can make. Choosing the wrong one for your caller is the most common failure.
```
TucksonDev: This introduction is a bit confusing (both the text and the table below). It's not 100% accurate, but in any case, we already have a guide that explains how to do the registration ([high-level explanation](https://docs.arbitrum.io/how-arbitrum-works/deep-dives/token-bridging#setting-up-your-token-with-the-generic-custom-gateway), [guide](https://docs.arbitrum.io/arbitrum-essentials/bridging/configure-token-gateway/generic-custom)).

I'd introduce the section saying something like "Registering a token or gateway involves sending specific transactions to the TokenBridge router or gateway. See this guide to understand how the process goes" or something like that.

TucksonDev: Additionally, there's a "FAQ" section in the guide. Would probably make sense to unify that type of content either in this page or there.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+| `registerTokenToL2`      | `L1CustomGateway` | `NOT_ARB_ENABLED` | The token contract itself, not an externally owned account |
+| `forceRegisterTokenToL2` | `L1CustomGateway` | `ONLY_OWNER`      | The gateway owner, normally the `UpgradeExecutor`          |
+| `setGateway`             | `L1GatewayRouter` | `NOT_ARB_ENABLED` | The token contract itself                                  |
+| `setGateways`            | `L1GatewayRouter` | `ONLY_OWNER`      | The router owner, normally the `UpgradeExecutor`           |
+
+### Wrong caller for the registration path
+
+`registerTokenToL2` and `setGateway` are designed to be called _by the token_. They revert with `NOT_ARB_ENABLED` unless the caller implements the <a data-quicklook-from="custom-arbtoken">custom token</a> interface and returns the expected value from `isArbitrumEnabled()`. Calling either from an externally owned account fails no matter how that account is funded or permissioned.
+
+For a token that can't be modified, use the owner-side path instead. `forceRegisterTokenToL2` takes parallel arrays and bypasses the token-side check:
```
TucksonDev:
```suggestion
For a token that can't be modified, use the chain-owner-side path instead. `forceRegisterTokenToL2` takes parallel arrays and bypasses the token-side check:
```

So it's not confused with being the owner of the token

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+```solidity
```
TucksonDev: This piece of code and the text below can be removed IMO. Any operator taking this action is probably familiar with the process.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+    address[] calldata _l2Addresses,
+    uint256 _maxGas,
+    uint256 _gasPriceBid,
+    uint256 _maxSubmissionCost
+) external payable virtual onlyOwner returns (uint256)
+```
+
+Because it's `onlyOwner` and the owner is the `UpgradeExecutor` on chains deployed by the creator, the call must be routed through the executor rather than sent from the <a data-quicklook-from="chain-owner">chain owner</a>'s account directly.
+
+### Underfunded retryable tickets
```
TucksonDev: This section can be removed, since it's a duplicate of the one below, and IMO that one is better explained.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+### Registration succeeded on one chain only
+
+Because registration is asynchronous, the parent chain transaction can succeed while the child chain retryable fails or expires unredeemed. The result is a half-applied registration: one router resolves your custom gateway, and the other falls back to its default. Deposits and withdrawals then disagree about which gateway owns the token.
+
+After any registration, confirm `getGateway` returns the same gateway on both routers. If the child chain side is missing, locate the retryable ticket from the parent chain transaction and redeem it. To learn how redemption works, see [Parent to child chain messaging](/how-arbitrum-works/deep-dives/l1-to-l2-messaging.mdx).
+
+### Array length and remapping errors
+
+- `INVALID_LENGTHS` from `forceRegisterTokenToL2` and `WRONG_LENGTH` from `setGateways` both mean the address arrays you passed have different lengths.
```
TucksonDev: This point can be removed. The other 3 are useful

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+- `NO_UPDATE_TO_DIFFERENT_ADDR` means the token already maps to a _different_ child chain address. Registration is not a general-purpose update mechanism—re-registering with the same address is accepted, but pointing an already-registered token at a new child chain address is rejected. Confirm the intended mapping before the first registration.
+- `NOT_TO_CONTRACT` means the gateway address you supplied has no code.
+- `TOKEN_NOT_HANDLED_BY_GATEWAY` means the gateway doesn't recognize the token you're pointing it at. This is an ordering problem: register the token with the gateway first, then register the gateway with the router.
+
+## A WETH gateway is missing
+
+Your chain has no <a data-quicklook-from="weth-gateway">WETH gateway</a>, no child chain `aeWETH`, and no **WETH** bridging path.
+
+If your chain uses a <a data-quicklook-from="custom-gas-token">custom gas token</a>, it was never deployed. In `createTokenBridge`, the entire **WETH** gateway block is conditional on `feeToken == address(0)`, so a custom gas token chain skips it. This matches step 5 of the deployment guide, which applies only to **ETH**-based chains.
```
TucksonDev:
```suggestion
If your chain uses a <a data-quicklook-from="custom-gas-token">custom gas token</a>, it doesn't need a WETH gateway since ETH is not the native token of the chain and doesn't require specific handling. In a custom gas token chain, WETH can be bridge through the default gateway.
```

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+On an **ETH**-based chain, the gateway is wired to the **WETH** address stored by the creator you deployed through, not to one you choose at chain creation time. To learn which creator you used, see [Canonical factory contracts](/launch-arbitrum-chain/deploy/canonical-factory-contracts.mdx).
+
+## Token bridge contracts show as unverified
```
TucksonDev: I'm not sure this issue is recent. A while ago, Blockscout needed specific verification for their contracts, but nowadays they already read verified code from a centralized source, so not sure this is still happening. Currently, whenever we deploy a new TokenBridgeCreator, we verify all contracts, so any TokenBridge created from there should automatically have their contracts verified.

In any case, the resolution here only describes how to use the script. If no recent reports have been received, I'd remove this section. If we want to maintain it for some reason, it might be better to just point to the doc in the repository, since that's where we would modify instructions if we ever change the script: https://github.com/OffchainLabs/token-bridge-contracts/blob/develop/docs/deployment.md#verify-orbit-contracts-source-code-on-the-blockscout

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+- `BASECHAIN_WETH`, the parent chain **WETH** address the creator stores.
+- `ROLLUP_ADDRESS`, read only when you leave `GAS_LIMIT_FOR_L2_FACTORY_DEPLOYMENT` unset.
+
+A _missing_ variable produces a clear error instead, such as `Missing BASECHAIN_WETH in env vars`. Seeing the ENS error means the variable is set but malformed.
+
+### Resolution
+
+Verify that each address is a bare, checksummed hex string: no quotes, no whitespace, `0x` prefixed, 42 characters. Placeholder text, a missing `0x` prefix, and a truncated address are the common defects.
+
+## `outboundTransfer` reverts or moves the wrong tokens
```
TucksonDev:
```suggestion
## Calling `outboundTransfer` directly on the child chain gateway reverts or moves the wrong tokens
```

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+Scripts that call `outboundTransfer` directly on a child chain gateway, bypassing the router, commonly fail. `L2ArbitrumGateway.outboundTransfer` branches on whether the caller is the router:
```
TucksonDev: I'd avoid using this like "commonly fail" here.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+if (isRouter(msg.sender)) {
+    (_from, _extraData) = GatewayMessageHandler.parseFromRouterToGateway(_data);
+} else {
+    _from = msg.sender;
+    _extraData = _data;
+}
+require(_extraData.length == 0, "EXTRA_DATA_DISABLED");
+```
+
+Direct callers take the `else` branch, which produces three distinct problems:
```
TucksonDev, line 115: There are multiple points here that are irrelevant, hard to reach or confusing.

I'd summarize this into:
- Direct calls must pass empty bytes as `_data`, otherwise the call reverts with `EXTRA_DATA_DISABLED`.
- With no router, `_from` becomes `msg.sender`. A contract calling the gateway on a user's behalf has its own balance debited, not the user's.
- `_l1Token` is always the **parent chain** token address, even when calling the child chain gateway. Passing the child chain token address resolves to the wrong token or no token at all.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+- **Tokens debited from the wrong address.** With no router, `_from` becomes `msg.sender`. A contract calling the gateway on a user's behalf has its own balance debited, not the user's. There's no parameter to override this—supplying the sender is precisely what the router does.
+
+Two further details are easy to miss:
+
+- `_l1Token` is always the **parent chain** token address, even when calling the child chain gateway. Passing the child chain token address resolves to the wrong token or no token at all.
+- `ONLY_COUNTERPART_GATEWAY` comes from `finalizeInboundTransfer`, which only accepts calls from the aliased counterpart gateway. It's not a function to call yourself.
+
+### Resolution
+
+Call `outboundTransfer` on `L2GatewayRouter` rather than on the gateway. The router resolves the correct gateway, encodes the sender, and passes the empty extra data the gateway requires. Use the four-parameter overload, which forwards to the six-parameter version with zeroed gas parameters.
```
TucksonDev: I'd modify this resolution section like so:

The recommended path is to transfer tokens through the TokenBridge router, not the gateway directly. If you're bypassing the router because it resolves the wrong gateway, verify whether the token has been correctly registered in the router. If it is (for example, because a new gateway has been deployed for that token), follow the recommendation above to send your transaction.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+If you're bypassing the router because it resolves the wrong gateway, the routing table is the actual problem. Fix the registration rather than working around it—see [Custom gateway registration fails](#custom-gateway-registration-fails).
+
+## Quick reference: deployment and verification
```
TucksonDev: Not sure these sections are useful, but if they are, please update them with the new changes

TucksonDev (changes requested):

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
pete-vielhaber: I'm glad you agree with the duplication; I thought this was repetitive. I feel it flows better being removed.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
TucksonDev: Ah, I think my comment might have been a bit confusing. I meant that the proposed text live in the "What causes this problem" section, while keeping the previous introduction for the overall "Bridged USDC routes to the standard gateway" section. The previous introduction was:

You deployed a bridged **USDC** gateway following [the bridged USDC standard](/launch-arbitrum-chain/integrations/bridged-usdc.mdx), but the SDK and bridge UI keep sending **USDC** through the standard **ERC-20** gateway, producing the default bridged token rather than the Circle-standard one.

docs/launch-arbitrum-chain/deploy/token-bridge-troubleshooting.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+1. **Query the router directly.** Call `getGateway` on `L1GatewayRouter` with your parent chain **USDC** address. If it returns the standard **ERC-20** gateway rather than your **USDC** gateway, registration never landed.
+2. **Execute `registerUsdcGatewayTx.json`.** Submit the stored calldata through your `UpgradeExecutor` using the account holding the executor role.
+3. **Check both sides.** Registration crosses chains via a <a data-quicklook-from="retryable-ticket">retryable ticket</a>, so confirm the child chain router resolves correctly too. See [Registration succeeded on one chain only](#registration-succeeded-on-one-chain-only).
+
+## Custom gateway registration fails
+
+Registering a token or gateway involves sending specific transactions to the `TokenBridge` router or gateway. This section will cover common troubleshooting, for a deeper understanding refer to the [Generic custom gateway](/arbitrum-essentials/bridging/configure-token-gateway/generic-custom.mdx) page.
+
+| Entry point              | Contract          | Guard             | Who can call it                                            |
```
TucksonDev: I'd remove this table, since it's a bit confusing, given that the context is below. Additionally, it doesn't provide a lot of value, IMO. The error codes are a good reference but, `NOT_ARB_ENABLED` is already mentioned below, and `ONLY_OWNER` is a known error (so any chain operator is aware of it)

TucksonDev (changes requested):
Looking great! Just a few pending comments and should be good to publish 💪

TucksonDev (approved):
LGTM 🚀

PR #3523, opened 2026-08-11
Eth send raw transaction conditional

anegg0:
## Description

New reference page for `eth_sendRawTransactionConditional`, plus two diagrams. `rpc-methods.mdx` now links to it instead of duplicating a table, and `sidebars.js` gains the entry.

## Document type

- [ ] Gentle introduction
- [ ] Quickstart
- [ ] How-to
- [ ] Concept
- [ ] FAQ
- [ ] Troubleshooting
- [x] Reference
- [ ] Third-party content
- [x] Codebase changes
- [ ] Not applicable

## Checklist

- [x] I have read the [CONTRIBUTE.md](../CONTRIBUTE.md) guidelines
- [x] My changes follow the style conventions outlined in CONTRIBUTE.md
- [x] I have used sentence-case for titles and headers
- [x] I have used descriptive link text (not "here" or "this")
- [x] I have separated procedural from conceptual content where appropriate
- [x] I have tested my changes locally with `yarn start` or `yarn build`
- [x] My code follows the existing code style and conventions
- [x] I have added/updated frontmatter for new documents
- [x] I have checked for broken links
- [ ] I have verified that my changes don't break the build
- Third-party docs only: Do you agree to the third-party content policy outlined within [CONTRIBUTE.md](../CONTRIBUTE.md)?
  - [ ] Yes
  - [x] Not applicable

## Additional Notes

Also fixes two pre-existing site-wide layout bugs found while testing the page:

- `VanillaAdmonition` did not support `type="caution"`, so those admonitions rendered with no style, title, or icon on 26 pages. Added the type and a `.caution` class.
- The desktop table of contents wrapped below the article when a code line was wider than the column. Flex items default to `min-width: auto`, so min-content overrode the column's `max-width`. Added `min-width: 0` to the article column and the main container.

The layout fixes are in shared files (`src/css/custom.scss`, `src/components/VanillaAdmonition/`) and change rendering on every page, not just this one. Verified in the browser at 1600px, 1280px, and 390px: the table of contents returns to the right rail, page-level horizontal scrolling is gone, `rpc-methods` renders identically to before, and full-bleed tables are unaffected. `configure-dac.mdx`, the worst pre-existing case, is also fixed.

Local verification covered `yarn start` and `yarn typecheck`; a full `yarn build` has not been run, so CI is the gate for link checking.


docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+| Check                                 | Metric prefix                                  |
+| ------------------------------------- | ---------------------------------------------- |
+| Prechecker, current state             | `arb/txprechecker/conditionaltx/currentstate/` |
+| Prechecker, state about 2 seconds old | `arb/txprechecker/conditionaltx/oldstate/`     |
+| Sequencer, block being built          | `arb/sequencer/conditionaltx/`                 |
+
+## Limits
+
+Nitro sets no cap on the number of accounts or storage slots you list in `knownAccounts`. Every listed slot costs the node a state read on each of the three checks, so a large `knownAccounts` map stays cheap for you and grows expensive for the node.
```
zk-Lumi, line 225: Is this something we should call out? How do we enforce DOS on this?

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+Nitro reads three different sources when it evaluates your conditions: the <a data-quicklook-from="parent-chain">parent chain</a> block number, the <a data-quicklook-from="child-chain">child chain</a> block timestamp, and child chain contract storage.
+
+| Field                                 | Type     | Your transaction stays valid only while…                  |
+| ------------------------------------- | -------- | --------------------------------------------------------- |
+| `knownAccounts`                       | `object` | Each listed account still holds the storage you describe. |
+| `blockNumberMin` and `blockNumberMax` | quantity | The parent chain block number sits inside the range.      |
+| `timestampMin` and `timestampMax`     | quantity | The child chain block timestamp sits inside the range.    |
+
+<ImageZoom src="/img/arb-conditional-tx-scope.svg" className="img-900px" caption="Each conditional option reads a different source." alt="Three boxes. blockNumberMin and blockNumberMax read the parent chain block number. timestampMin and timestampMax read the child chain block timestamp. knownAccounts reads child chain contract storage." />
```
zk-Lumi, line 75: Its a little confusing the order of these is different than the table above

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+List only the slots your transaction depends on. For an ERC-4337 bundle, that usually means the nonce and deposit slots of each sender, rather than the storage root of each account.
+
+## Differences from other implementations
+
+Other stacks expose a method with the same name and a different contract. If you port a bundler or relay to an <a data-quicklook-from="arbitrum-chain">Arbitrum chain</a>, check these four points.
+
+| Behavior                              | On Arbitrum chains                                                                                                |
+| ------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
+| `blockNumberMin` and `blockNumberMax` | Read the parent chain block number. Other stacks read their own chain's block number.                             |
+| Old-state re-check                    | Nitro also evaluates your conditions against state about 2 seconds old. Other stacks evaluate current state only. |
```
zk-Lumi, line 242: What other stacks? Are we positive about this?

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+List only the slots your transaction depends on. For an ERC-4337 bundle, that usually means the nonce and deposit slots of each sender, rather than the storage root of each account.
+
+## Differences from other implementations
+
+Other stacks expose a method with the same name and a different contract. If you port a bundler or relay to an <a data-quicklook-from="arbitrum-chain">Arbitrum chain</a>, check these four points.
+
+| Behavior                              | On Arbitrum chains                                                                                                |
+| ------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
+| `blockNumberMin` and `blockNumberMax` | Read the parent chain block number. Other stacks read their own chain's block number.                             |
```
zk-Lumi, line 241: What other stacks? Are we positive about this?

zk-Lumi: "may" could be a good way to be safe

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
zk-Lumi: "may" could be a good way to be safe

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+This method is not part of the standard Ethereum JSON-RPC API. Nitro implements it, and so do several other stacks, but the semantics differ between them. If you port a client from another stack, read [Differences from other implementations](#differences-from-other-implementations) first.
+
+</VanillaAdmonition>
+
+## When to use it
+
+Use `eth_sendRawTransactionConditional` when your transaction makes sense only against a specific view of <a data-quicklook-from="chain-state">chain state</a>, and you would rather send nothing than send a transaction that reverts.
+
+[ERC-4337](https://eips.ethereum.org/EIPS/eip-4337) account abstraction is the original motivation. A bundler packs many user operations into one transaction. Between the moment the bundler simulates the bundle and the moment the sequencer executes it, another transaction can change a nonce or a balance that the bundle depends on. The bundle then reverts, and the bundler pays the gas. When you list the storage slots the bundle depends on, that failure moves from "executed and reverted" to "rejected before execution".
```
zk-Lumi, line 28: This section is kind of confusing and seems unnecessary context - What does 4337 have to do with `eth_sendRawTransactionConditional` / how did it motivate it, and to whom?

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+description: 'Reference for eth_sendRawTransactionConditional on Arbitrum chains: the conditions you attach to a transaction, where Nitro evaluates them, the error codes it returns, and the node flags that control the checks.'
+user_story: 'As a bundler or relay operator, I want to attach preconditions to a transaction so that the sequencer drops it instead of including a transaction that reverts against changed state.'
+target_audience: 'developers who submit transactions programmatically, such as ERC-4337 bundler and relay operators'
+content_type: reference
+author: anegg0
+sme: zk-lumi
+displayed_sidebar: arbitrumEssentialsSidebar
+---
+
+`eth_sendRawTransactionConditional` submits a signed <a data-quicklook-from="transaction">transaction</a> together with a set of preconditions. Nitro nodes evaluate those preconditions before the <a data-quicklook-from="sequencer">sequencer</a> includes the transaction. If any precondition fails, the node rejects the transaction and never includes it, so you pay nothing.
```
zk-Lumi, line 14: We should probably mention our implementation is based on [ERC-7796](https://eips.ethereum.org/EIPS/eip-7796)

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+
+<ImageZoom src="/img/arb-conditional-tx-checkpoints.svg" className="img-900px" caption="Nitro evaluates conditional transaction conditions at three points before inclusion." alt="A vertical flow. Your app signs the transaction and attaches the conditions, then sends it to a prechecker node. The prechecker runs check 1 against current state and check 2 against state about two seconds old. The sequencer runs check 3 against the block it is building, then includes the transaction in a child chain block. If any check fails, the node rejects the transaction with JSON-RPC error -32003 and never includes it." />
+
+1. **The <a data-quicklook-from="prechecker-node">prechecker node</a> evaluates current state.** The node you submit to evaluates your conditions against its latest block before it forwards anything.
+
+2. **The prechecker evaluates older state.** The node then walks back to a block that is at least `required-state-age` seconds old, which defaults to 2 seconds, and evaluates the same conditions against that older state. A failure here returns the original message behind the prefix `conditions check failed for old state`.
```
zk-Lumi, line 117: According to Claude (should be verified by Nitro): 

> The actual loop in tx_pre_checker.go walks back until either the state is 2 seconds old or it has walked required-state-max-blocks blocks (default 4), whichever comes first. 
> - On Arbitrum One with ~250ms blocks, defaults mean the old state is typically only ~4 blocks (~1 second) old, not 2 seconds. 
> - And on a quiet chain where the latest block is already 2+ seconds old, the walk-back lands on the same header and the second check is skipped entirely - so a condition that became true in the latest block passes. Neither direction of the "last 2 seconds" claim is a guarantee. 
> 
> Suggested framing: "walks back up to required-state-max-blocks blocks looking for state at least required-state-age seconds old, and re-checks against whatever it lands on - with defaults on Arbitrum One that's roughly the last ~1 second; when the head block is already old enough, the second check is skipped."

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+## Parameters
+
+| Position | Type     | Description                                                                               |
+| -------- | -------- | ----------------------------------------------------------------------------------------- |
+| 1        | `string` | The signed transaction, RLP-encoded and hex-prefixed. Same as `eth_sendRawTransaction`.   |
+| 2        | `object` | The conditions. Every field is optional, and the node evaluates only the fields you send. |
+
+The method returns the transaction hash, exactly as `eth_sendRawTransaction` does.
+
+Three transaction kinds never reach the condition checks. When you send a blob transaction, an Arbitrum internal transaction type, or a transaction without [EIP-155](https://eips.ethereum.org/EIPS/eip-155) replay protection, Nitro rejects it first.
```
zk-Lumi, line 63:
```suggestion
Three transaction kinds never reach the condition checks. When you send a blob transaction, an Arbitrum internal transaction type, or a transaction without [EIP-155](https://eips.ethereum.org/EIPS/eip-155) replay protection, a Nitro node running default configuration rejects it before it evaluates your conditions.
```

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+## When to use it
+
+Use `eth_sendRawTransactionConditional` when your transaction makes sense only against a specific view of <a data-quicklook-from="chain-state">chain state</a>, and you would rather send nothing than send a transaction that reverts.
+
+[ERC-4337](https://eips.ethereum.org/EIPS/eip-4337) account abstraction is the original motivation. A bundler packs many user operations into one transaction. Between the moment the bundler simulates the bundle and the moment the sequencer executes it, another transaction can change a nonce or a balance that the bundle depends on. The bundle then reverts, and the bundler pays the gas. When you list the storage slots the bundle depends on, that failure moves from "executed and reverted" to "rejected before execution".
+
+The same pattern helps any submitter that pays gas on behalf of other people:
+
+- Relays that sponsor meta-transactions.
+- Liquidation and arbitrage bots that must not act on a stale price.
```
tsahee, line 33: it's designed not to be usable for liquidation and arb-bots

docs/arbitrum-essentials/arbitrum-vs-ethereum/send-raw-transaction-conditional.mdx
```diff
@@ -0,0 +1,252 @@
...
+target_audience: 'developers who submit transactions programmatically, such as ERC-4337 bundler and relay operators'
+content_type: reference
+author: anegg0
+sme: zk-lumi
+displayed_sidebar: arbitrumEssentialsSidebar
+---
+
+`eth_sendRawTransactionConditional` submits a signed <a data-quicklook-from="transaction">transaction</a> together with a set of preconditions. Nitro nodes evaluate those preconditions before the <a data-quicklook-from="sequencer">sequencer</a> includes the transaction. If any precondition fails, the node rejects the transaction and never includes it, so you pay nothing.
+
+The method extends `eth_sendRawTransaction`. It takes the same signed transaction payload and adds a second parameter that describes the state your transaction depends on.
```
tsahee, line 16: eth_sendRawTransactionConditional can cause a large burden to the system. To avoid it we add a 1 second delay to any transaction incoming. The condition is measured twice - before and after the delay and must be true both times. It is designed to help actors (mostly AA bundlers) make sure nothing on-chain rugs them, but it cannot be used to detect short-lived on-chain opportunities.

zk-Lumi:
Initial purpose of this PR:
> It allows users to specify a max timestamp with ms precision or conditional transactions based on current contract state. We keep having market makers that aren't aware of it, and when you google it OP and Conduit explanations come up

Feedback from Platform team
> sendRawTransactionConditional is very much not intended for market makers, but for AA bundlers. It's a huge DOS vectors by MEV searchers, so to make sure it's not exploited we added in a 1 second delay to any transaction that comes in. We check condition twice: once when the tx comes in, and once a second later, right before we include it. It's designed to be unusable for exploiting on-chain opportunities.

> Any public documentation we present must press this point. On it's face a conditional sendrawtransaction seems like MEV-bot heaven - exploit opportunities by spamming without paying gas for when the opportunity doesn't exist. We must stress that it's not the case and make sure people don't try using it that way.

PR #3542, opened 2026-08-20
Remove 51 redirects to the deleted /sdk section

anegg0:
`docs/sdk` does not exist. `sidebars.js` links the SDK to
`github.com/OffchainLabs/arbitrum-sdk` instead. The 51 `/sdk-docs/*` redirects
still resolved to `/sdk/*`, so an old link took a redirect hop and then 404'd.
Removing them returns a plain 404.

All 51 sat outside the `__redirects-autogen__` sentinels, so `yarn sync-redirects`
will not restore them. Redirects that merely contain "sdk" in the name are
untouched; each resolves to a real page.

639 -> 588 redirects. Deletions only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pete-vielhaber (approved):
LGTM

